PR #9, opened 2025-03-13
Update network-services-operator with multicluster support, migrate e2e tests to Chainsaw

joshlreese:
## Multicluster support

This PR introduces the use of the (currently experimental) [multicluster-runtime](https://github.com/multicluster-runtime/multicluster-runtime) library that brings the ability to attach an operator to multiple clusters with minimal changes to reconciler code. A new `cluster-discovery-mode` flag has been introduced, and supports the following options:

- **single**: Default behavior. The operator will behave as it does today, attaching to a single cluster.
- **kind**: The operator will leverage the `kind` library to discover kind clusters, and engage each one. I expect this to be useful in e2e tests as we move forward.
- **datum**: The operator will list `Project` resources in the API server that the default kubeconfig interacts with, and engages a cluster for each `Project` found. Currently it is expected that the project control plane is available via the path `/apis/resourcemanager.datumapis.com/v1alpha/projects/{project_name}/control-plane` on the same host that the main kubeconfig attaches to.

An initial implementation of a multicluster-runtime `Provider` for Datum which watches `Project` resources is included in this PR. As that library stabilizes, we'll move it out to another location that other Datum operators can use.

> [!NOTE]
> The multicluster-runtime dependency is currently referencing a fork. As changes are merged
> in to multicluster-runtime, such as https://github.com/multicluster-runtime/multicluster-runtime/pull/18, the dependency on the fork will be removed.

## End to end tests

End to end tests have been replaced with [Chainsaw](https://kyverno.github.io/chainsaw/latest/). I referenced the [opentelemetry-operator](https://github.com/open-telemetry/opentelemetry-operator) project's use of chainsaw in the Makefile and GitHub action changes.


internal/providers/datum/provider.go
```diff
@@ -0,0 +1,236 @@
...
+	// create cluster.
+	cl, err := cluster.New(cfg, p.opts.ClusterOptions...)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to create cluster: %w", err)
+	}
+	// for _, idx := range p.indexers {
+	// 	if err := cl.GetCache().IndexField(ctx, idx.object, idx.field, idx.extractValue); err != nil {
+	// 		return ctrl.Result{}, fmt.Errorf("failed to index field %q: %w", idx.field, err)
+	// 	}
+	// }
```
scotwells, line 178: Is this supposed to be commented out?

scotwells (approved):
Left one minor comment. Overall looks like this will work great. Love the documentation that's automatically generated off of the chainsaw tests. 🚀

internal/providers/datum/provider.go
joshlreese: I have it commented out since I haven't dug into why it's needed yet (I used the cluster-api provider as a reference). I suspect it's for being able to define field indexers at the multicluster.Manager level, but wanted to test before bringing in the code.

scotwells (approved):

PR #12, opened 2025-03-19
Datum gateways

joshlreese:
Draft PR - relates to https://github.com/datum-cloud/enhancements/issues/6

Will update the PR description before ready for review. I'll also be reviewing if it's appropriate to check in the charts for the kustomize apps like I have (kustomize pulls them down when executing, so could save some CI time).


scotwells (dismissed):

cmd/main.go
```diff
@@ -273,12 +261,9 @@ func main() {
...
 
-	if singleCluster != nil {
-		setupLog.Info("starting cluster for single cluster provider")
-		g.Go(func() error {
-			return ignoreCanceled(singleCluster.Start(ctx))
-		})
-	}
+	g.Go(func() error {
+		return ignoreCanceled(downstreamCluster.Start(ctx))
+	})
```
scotwells, line 266: Does it matter that `downstreamCluster.Start()` is going to be called twice since `downstreamCluster` is also being added to `runnables` in `initializeClusterDiscovery`?

joshlreese: It's actually the `deploymentCluster` that's passed into `initializeClusterDiscovery` and added to `runnables`, which is only used inside that function when the discovery mode is `single`.

scotwells (approved):

PR #14, opened 2025-04-02
Validate and publish kustomize manifests

OscarLlamas6:
### Overview
We need to validate that the Kustomize manifests defined in the repo work correctly for deploying the network services operator since we will now be using FluxCD to deploy the component for all of the project control planes.

### Acceptance Criteria

Validate that the kustomize manifests can be used to create all of the necessary resources [created by the project control plane today for the network-services-operator deployment](https://github.com/datum-cloud/project-control-plane-operator/blob/main/internal/controller/projectcontrolplane_controller.go#L892).

[Flux](https://fluxcd.io/flux/cmd/flux_push/) is used to published the kustomize manifests to an OCI repository with the same tagging conventions used by the container image used for the operator binary.

### Note

The Kustomize manifests we expose in this repo should only be ones that are used to deploy the application in a way that can also be used by the open source community. Any additional manifests (e.g. kubeconfig) that we use to deploy the operator in our own specific way should be kept in datum-infra.


.github/workflows/publish.yaml
```diff
@@ -11,3 +11,21 @@ jobs:
...
+      id-token: write
+      contents: read
+      packages: write
+    strategy:
+      matrix:
+        bundles:
+          - name: ghcr.io/datum-cloud/network-services-operator/network-services-crds
+            path: config/crd
+          - name: ghcr.io/datum-cloud/network-services-operator/network-services-operator
+            path: config/network-services-operator
```
scotwells, line 26: @OscarLlamas6 you should be able to use the `config/default` program to export the operator's manifests. 

```suggestion
          - name: ghcr.io/datum-cloud/network-services-operator-kustomize
            path: config/default
```

config/network-services-operator/deployment.yaml
scotwells, line 1: @OscarLlamas6 we should be able to leverage `config/default` since that includes `config/manager` which contains the deployment manifests for the operator.

.github/workflows/publish.yaml
OscarLlamas6: @scotwells what do you think about this to have both under */network-services-operator* ?

```css
bundles:
          - name: ghcr.io/datum-cloud/network-services-operator/network-services-crds
            path: config/crd
          - name: ghcr.io/datum-cloud/network-services-operator/network-services-operator-kustomize
            path: config/default
```

OscarLlamas6: @scotwells if we have config/default as a kustomization we dont need the **_ghcr.io/datum-cloud/network-services-operator/network-services-crds_** or do we need it? Since the /config/default kustomization includes crds it will install everything right?

PR #15, opened 2025-05-07
Add support for permitting specific clusters to use custom domains via a suffix match,

joshlreese:
As an interim step to support custom hostnames, this PR introduces the ability to configure hostname suffixes that may be used on a per-cluster basis via the configuration file. See `config/dev/config.yaml` and `config/e2e/config.yaml` for examples of how this is expected to work.

> [!IMPORTANT]
> Custom hostnames will not be wired into the `Status.Addresses` field of the Gateway, as the system is not responsible for ensuring DNS for those hostnames point at the correct location.

Other notable changes:

- Cleaned up e2e test suite to no longer rely on external systems for backends and removed duplicate tests.
- Corrected a bug where the observed generation was not being populated in the GatewayClass' Accepted condition.
- Corrected a bug that would result in the gateway controller processing deletions for non-applicable gateways.
- Corrected validation logic to only permit routes from the same namespace to attach to gateways.


internal/controller/gateway_controller.go
```diff
@@ -158,17 +162,19 @@ func (r *GatewayReconciler) ensureDownstreamGateway(
 	}
 	upstreamGatewayClassControllerName := string(upstreamGatewayClass.Spec.ControllerName)
 
-	// TODO(jreese) handle hostnames defined on upstream listeners
-	hostnames := []string{
+	// addressHostnames are default hostnames that are unique to each gateway, and
+	// will have DNS records created for them. Any custom hostnames provided in
+	// listeners WILL NOT be added to the addresses list in the gateway status.
+	addressHostnames := []string{
```
scotwells, line 168: Just to confirm, we're not adding them here because we don't want external DNS to try and create the DNS record in KnotDNS?

scotwells (approved):

internal/controller/gateway_controller.go
scotwells: Do you think it'll be confusing from a user's perspective that they don't see any status information for custom hostnames they added? Maybe those should be represented in a separate status field? More of a forward looking question, don't need to address it now since this is being done as a workaround.

joshlreese: > Do you think it'll be confusing from a user's perspective that they don't see any status information for custom hostnames they added?

I don't think it would be surprising for users familiar with the Gateway API. It's also important to remember that a hostname doesn't even need to be defined on the listeners, and we're auto-injecting ones that _just happen_ to align with the network addresses we allocate. Also, hostnames can be defined directly on routes (though we're prohibiting that today).

> Maybe those should be represented in a separate status field?

I'm not seeing one in the spec that would be appropriate for this information. It could be shoved into a message on a listener condition, but that feels unnecessary.

> More of a forward looking question, don't need to address it now since this is being done as a workaround.

This isn't a workaround - only the allow list is.

scotwells: All good points 👍

PR #17, opened 2025-05-13
Program downstream gateways to use a separate secret for each listener certificate

joshlreese:
@scotwells pointed out that cert-manager will correctly create a `Certificate` for each unique secret referenced by a `CertificateRef` on gateway listeners, and only include the hostname defined on that listener. This will address the OverlappingTLSConfig issue observed when evaluating the latest release of Envoy Gateway - details can be found at https://github.com/envoyproxy/gateway/issues/2675.

As part of this, I've introduced a utility that provides helpers for obtaining resource names for various length limits. If an input name exceeds the length limit, it will be truncated and suffixed with a hash of the full name. This is done for ease of development - in this particular scenario, a Gateway's name can be up to 253 characters, as can the listener's name.

I also made a minor tweak to finalize logic to prevent attempting to remove a finalizer that doesn't exist, which usually leads to object not found errors as the resource has already been removed.

### DIFF
--- a/internal/controller/gateway_controller.go
+++ b/internal/controller/gateway_controller.go
@@ -34,6 +34,7 @@ import (
 
 	"go.datum.net/network-services-operator/internal/config"
 	downstreamclient "go.datum.net/network-services-operator/internal/downstreamclient"
+	"go.datum.net/network-services-operator/internal/util/resourcename"
 )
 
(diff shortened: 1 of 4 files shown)


scotwells (approved):

joshlreese:
Example of a certificate before this change - note the SAN list:
image

After:
image

OscarLlamas6 (approved):

scotwells (approved):

PR #23, opened 2025-06-09
Adding Domain CRD

### LINKED ISSUE #21: Add Domain CRD to Manage Domain Resources
Define and register a Domain custom resource definition (CRD) to represent managed domain names in the Kubernetes API. This is the foundation for tracking domain ownership, verification status, and eventually health/WHOIS metadata.

### Acceptance Criteria
* A Domain CRD is created under apiVersion: datumapis.com/v1alpha
* Spec includes domain name (metadata.name) and any future-proof fields (e.g., labels, annotations)
* CRD is registered via controller and visible in the cluster
* CRD validation schema supports future extensibility
* Proper OpenAPI schema and validation for fields

REF: [Enhancement for Domains and DNS at Datum](https://github.com/datum-cloud/enhancements/pull/90)

mustela:
Define and register a Domain custom resource definition (CRD) to represent managed domain names in the Kubernetes API. This is the foundation for tracking domain ownership, verification status, and eventually health/WHOIS metadata.

Fixes #21


api/v1alpha/domain_types.go
```diff
@@ -0,0 +1,72 @@
...
+
+// DomainRegistrarStatus represents the registrar information for a domain
+type DomainRegistrarStatus struct {
+	IANAID         string   `json:"ianaID,omitempty"`
+	IANAname       string   `json:"ianaName,omitempty"`
+	CreatedDate    string   `json:"createdDate,omitempty"`
+	ModifiedDate   string   `json:"modifiedDate,omitempty"`
+	ExpirationDate string   `json:"expirationDate,omitempty"`
+	Nameservers    []string `json:"nameservers,omitempty"`
+	DNSSEC         string   `json:"dnssec,omitempty"`
```
joshlreese, line 49: This wasn't discussed in the enhancement, but maybe it'd make sense to have a `DNSSECStatus` struct with a `Signed bool` field, to keep the door open for more DNSSEC related checks? I don't have much domain knowledge here (no pun intended), so asked o3 for some input on what could be added in the future:

### Why wrap DNSSEC info in its own struct?

* **Forward-compatible** – adding new fields later won’t break clients already deserialising the CRD.
* **Clear boundary** – keeps DNSSEC data separate from registrar / WHOIS facts.
* **Optional inclusion** – the whole block can be omitted for TLDs that don’t support DNSSEC.

---

### Sketch of a `DNSSECStatus` struct

```go
type DNSSECStatus struct {
    // True if a DS record is present at the parent; false means the
    // delegation is insecure or unsigned.
    Signed bool `json:"signed"`

    // Short reason string (OK | DS_MISMATCH | NO_DS | VALIDATION_ERROR ...).
    DelegationStatus string `json:"delegationStatus,omitempty"`

    // Number of DS records found at the parent.
    DSCount int `json:"dsCount,omitempty"`

    // List of key tags present in those DS records.
    KeyTags []uint16 `json:"keyTags,omitempty"`

    // List of algorithm identifiers in use (RFC 8624 IANA table values).
    Algorithms []uint8 `json:"algorithms,omitempty"`

    // Earliest RRSIG expiration seen on apex DNSKEY (seconds from now).
    // Lets you warn about keys nearing expiry.
    EarliestRRSIGExpiry metav1.Time `json:"earliestRRSIGExpiry,omitempty"`

    // When the controller last performed a full DNSSEC check.
    LastChecked metav1.Time `json:"lastChecked"`
}
```

---

### How it would look in YAML

```yaml
status:
  dnssec:
    signed: true
    delegationStatus: OK
    dsCount: 2
    keyTags: [12345, 56789]
    algorithms: [8]           # RSASHA256
    earliestRRSIGExpiry: "2025-07-28T13:04:05Z"
    lastChecked: "2025-06-13T17:22:31Z"
```

---

### Why each field is useful

* **`DelegationStatus`** – surfaces common failure modes (DS present but doesn’t match DNSKEY, parent hasn’t published DS yet, etc.).
* **`KeyTags` / `Algorithms`** – quick visibility for ops when rolling keys or auditing algorithm policy.
* **`EarliestRRSIGExpiry`** – lets higher-level automation or alerts warn that the zone isn’t re-signing.
* **`LastChecked`** – makes it obvious if the reconciling controller is stalled or rate-limited.

If you only need a go/no-go indicator today, start with `Signed` and `DelegationStatus`. The struct leaves room to append the richer telemetry later without a breaking API change.

joshlreese (changes requested):

mustela:
@joshlreese updated!

api/v1alpha/domain_types.go
joshlreese: Moving from a `string` to a `struct` would be a breaking change, considering that I think  we should start with a struct with a single `Signed bool` field to start with.

joshlreese: @mustela - I reviewed the rest of your adjustments, I think this is the only remaining item I think we should address before moving forward.

mustela:
@joshlreese updated!

joshlreese (approved):

PR #26, opened 2025-06-25
Adding printer columns

mustela:
Adding "Print columns" to the Domain, Locations and Network types.

### DIFF
--- a/api/v1alpha/location_types.go
+++ b/api/v1alpha/location_types.go
@@ -71,7 +71,7 @@ type LocationStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Class",type="string",JSONPath=".spec.locationClassName"
-// +kubebuilder:printcolumn:name="City",type="string",JSONPath=".spec.topology['topology.datum.net/city-code']"
+// +kubebuilder:printcolumn:name="City",type="string",JSONPath=`.spec.topology.topology\.datum\.net/city-code`
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=`.status.conditions[?(@.type==\"Ready\")].status`
 // +kubebuilder:printcolumn:name="Reason",type="string",JSONPath=`.status.conditions[?(@.type==\"Ready\")].reason`


api/v1alpha/location_types.go
```diff
@@ -70,6 +70,13 @@ type LocationStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Class",type="string",JSONPath=".spec.locationClassName"
+// +kubebuilder:printcolumn:name="Provider",type="string",JSONPath=".spec.provider.gcp.projectId"
+// +kubebuilder:printcolumn:name="Region",type="string",JSONPath=".spec.provider.gcp.region"
+// +kubebuilder:printcolumn:name="Zone",type="string",JSONPath=".spec.provider.gcp.zone"
```
scotwells, line 76: These are going to become inaccurate with the upcoming AWS provider we offer. Maybe we prefix the columns with "GCP"? Hoping we can leverage CEL expressions in the next Kubernetes release to have more control over the values in these fields.

api/v1alpha/domain_types.go
```diff
@@ -8,6 +8,13 @@ import (
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Domain",type="string",JSONPath=".spec.domainName"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+// +kubebuilder:printcolumn:name="Reason",type="string",JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
+// +kubebuilder:printcolumn:name="Registrar",type="string",JSONPath=".status.registrar.ianaName"
+// +kubebuilder:printcolumn:name="DNSSEC",type="boolean",JSONPath=".status.registrar.dnssec.signed"
+// +kubebuilder:printcolumn:name="Expires",type="date",JSONPath=".status.registrar.expirationDate"
```
scotwells, line 17: Should we add a printer column for the DNS verification record? This would make it easy for a user to copy the value from the CLI when configuring a domain.

api/v1alpha/location_types.go
joshlreese: Yes, they will become inaccurate. I’d recommend that we do not add provider specific output to the printer columns. It’s actually very likely that the provider field is deprecated by a new field that will be a reference to a `ProviderLink` or a `controller`.

Instead, we should add well known topology keys like `topology.datum.net/city-code`, which carries the values users will want to target in their workload placement settings.

api/v1alpha/location_types.go
```diff
@@ -70,6 +70,11 @@ type LocationStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Class",type="string",JSONPath=".spec.locationClassName"
+// +kubebuilder:printcolumn:name="City",type="string",JSONPath=".spec.topology['topology.datum.net/city-code']"
```
joshlreese, line 74: This format doesn't seem to work for me locally - is it working for you?

The following does work:

```suggestion
// +kubebuilder:printcolumn:name="City",type="string",JSONPath=`.spec.topology.topology\.datum\.net/city-code`
```

mustela: You are right. My mistake was not actually checking the value. I thought that if it didn’t fail to build or install the CRD, the column would just work. 

Before:
```
kubectl get locations
NAME            CLASS           CITY   AGE
us-central1-a   datum-managed          24d
```

After: 

```
NAME            CLASS           CITY          AGE
us-central1-a   datum-managed   us-central1   24d
```

joshlreese (approved):

PR #32, opened 2025-07-12
feat: New HTTPProxy CRD to use as a convenient method to set up reverse proxies.

### LINKED ISSUE #30: Define HTTPProxy CRD
See https://github.com/datum-cloud/enhancements/issues/200#issuecomment-3057610827

joshlreese:
See https://github.com/datum-cloud/enhancements/issues/200 for details on the goals of this new API type.

This type builds on top of existing Gateway API structures where they do not require or support referencing external entities. These include route matches, filters, addresses, and various constants.

See the contents of the following content to see what YAML manifests would look like:

https://github.com/datum-cloud/enhancements/issues/200#issuecomment-3057610827

> [!NOTE]
> Significant documentation and validation rules are copied from the Gateway API spec.

Fixes #30

### DIFF
--- a/api/v1alpha/httpproxy_types.go
+++ b/api/v1alpha/httpproxy_types.go
@@ -63,7 +63,7 @@ type HTTPProxyRule struct {
 	// sent.
 	//
 	// +kubebuilder:validation:MinItems=1
-	// +kubebuilder:validation:MaxItems=16
+	// +kubebuilder:validation:MaxItems=1
 	Backends []HTTPRouteBackend `json:"backends,omitempty"`
 }
 
--- a/config/crd/bases/networking.datumapis.com_httpproxies.yaml
+++ b/config/crd/bases/networking.datumapis.com_httpproxies.yaml
@@ -878,7 +878,7 @@ spec:
                         required:
                         - endpoint
                         type: object
-                      maxItems: 16
+                      maxItems: 1
                       minItems: 1
                       type: array
                     filters:


mustela (approved):
LGTM!

api/v1alpha/httpproxy_types.go
```diff
@@ -0,0 +1,140 @@
...
+	// +kubebuilder:validation:XValidation:message="RequestRedirect filter cannot be repeated",rule="self.filter(f, f.type == 'RequestRedirect').size() <= 1"
+	// +kubebuilder:validation:XValidation:message="URLRewrite filter cannot be repeated",rule="self.filter(f, f.type == 'URLRewrite').size() <= 1"
+	Filters []gatewayv1.HTTPRouteFilter `json:"filters,omitempty"`
+
+	// Backends defines the backend(s) where matching requests should be
+	// sent.
+	//
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
+	Backends []HTTPRouteBackend `json:"backends,omitempty"`
```
joshlreese, line 67: @scotwells / @mustela - I'd like to limit the number of backends a rule can have to 1 for now, as I want to be able to use the `URLRewrite` filter for path rewriting, and Envoy Gateway does not currently support GA filters other than `RequestHeaderModifier` and `ResponseHeaderModifier` in a BackendRef. To use the URLRewrite filter, I'll need to add it into the rule's filter list, and allowing multiple backends would introduce issues as their host values may differ.

I'm thinking we can keep the type as a slice and limit the number of values via validation by setting `+kubebuilder:validation:MaxItems=1`, so we can lift the limitation when possible.

What do you think?

See https://github.com/envoyproxy/gateway/blob/64f3576cffa892acdf3af396b51b29b7a28cd4cc/internal/gatewayapi/validate.go#L122 for the Envoy Gateway code that enforces what filters are allowed.

joshlreese: > I'm going to test using the `RequestHeaderModifier` filter to manipulate the host header. If that works, I'll adjust the spec to allow for more backends again.

Per https://github.com/envoyproxy/gateway/blob/64f3576cffa892acdf3af396b51b29b7a28cd4cc/internal/gatewayapi/filters.go#L553

> Envoy does not allow modification the pseudo headers and the host header

So, I'll leave the number of backends at 1 per rule for now and leverage the URLRewrite filter.

PR #34, opened 2025-07-15
feat: Add HTTPProxy controller and webhook implementation

### LINKED ISSUE #31: Create HTTPProxy controller
Create a controller to reconcile HTTPProxy resources within Datum Projects.

This controller MUST:

- Observe HTTPProxy resources in Datum Projects (upstream clusters in the context of the operator).
- Maintain a Gateway that aligns 1:1 with the HTTPProxy
  - MUST have `gatewayClassName` set to `datum-external-global-proxy`.
  - MUST have a listener for port 80.
  - MUST have a listener for port 443, TLS mode set to Terminate, and the option `gateway.networking.datumapis.com/certificate-issuer` set to `auto`.
  - MUST have `allowedRoutes.namespaces.from` set to `Same`
- Process HTTPProxy rules and maintain an HTTPRoute for each, carrying over filters.
  - Process the `backends` defined for each rule
    - For each backend
      - Extract upstream host information such as appProtocol, port, and hostname from the endpoint URL.
      - Maintain an `EndpointSlice` with a `FQDN` AddressType and a single address with the host portion of the URL, and port information determined from the URL.
        - Note, the FQDN AddressType has been deprecated, but as we control the programming of an HTTPProxy, we can easily transition to an alternative once we have one. This is in an effort to not block MVP goals.
      - When the host portion of the URL is a DNS address, place the value in a URLRewrite filter on the BackendRef
      - When a path exists in the endpoint URL, ensure a rewrite filter is in place to add the path to incoming requests for the BackendRef.
      - Add an entry to the BackendRefs for the rule, referencing the EndpointSlice at the port determined from the URL.

joshlreese:
## Summary

- **New HTTPProxy CRD**: Simplified interface for creating reverse proxies with
  automatic Gateway API resource management
- **Controller implementation**: Manages Gateway, HTTPRoute, and EndpointSlice
  resources in upstream control planes based on HTTPProxy specifications
- **Validation webhook**: Ensures HTTPProxy configurations are valid before
  creation/updates

## Key Features

- **Automatic resource generation**: Creates and manages Gateway API resources
  (Gateway, HTTPRoute, EndpointSlice) from HTTPProxy specs
- **Backend validation**: Ensures backend URLs are valid and properly formatted
- **Status management**: Provides detailed status conditions for troubleshooting

Fixes #31


internal/controller/httpproxy_controller.go
```diff
@@ -0,0 +1,452 @@
...
+
+		gateway.Spec = desiredResources.gateway.Spec
+
+		return nil
+	})
+	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			programmedCondition.Status = metav1.ConditionFalse
+			programmedCondition.Reason = networkingv1alpha.HTTPProxyReasonConflict
+			programmedCondition.Message = fmt.Sprintf("Underlying Gateway with the name %q already exists and is owned by a different resource.", gateway.Name)
```
scotwells, line 134: Should we instruct the user to recreate the HTTP Proxy with a different name in this scenario? Might not be immediately obvious to them how to resolve the issue.

Same question for other resources that will have a similar status message.

scotwells (approved):
Super clean implementation. Only a minor comment about the error message that's returned in the status condition to give more direct guidance to the user on how to resolve the issue.

internal/controller/httpproxy_controller.go
```diff
@@ -0,0 +1,452 @@
...
+							Hostname: ptr.To(gatewayv1.PreciseHostname(host)),
+						},
+					})
+				}
+			}
+
+			endpointSlice := &discoveryv1.EndpointSlice{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: httpProxy.Namespace,
+					Name:      fmt.Sprintf("%s-%d-%d", httpProxy.Name, ruleIndex, backendIndex),
```
zachsmith1, line 389: Using GPT to review some of this:

Your naming scheme fmt.Sprintf("%s-%d-%d", httpProxy.Name, ruleIndex, backendIndex) could cause conflicts if:

HTTPProxy names are long (Kubernetes has 63-character limits)
Multiple HTTPProxies have similar names
Rule/backend indices create overlapping names

Recommendation: Use a hash-based approach or UUID suffix for uniqueness.

internal/validation/httpproxy_validation.go
```diff
@@ -0,0 +1,102 @@
...
+		if u.User != nil {
+			allErrs = append(allErrs, field.Invalid(endpointFieldPath.Key("userinfo"), fmt.Sprintf("%s:redacted", u.User.Username()), "endpoint must not have a userinfo component"))
+		}
+
+		for _, msg := range validation.IsDNS1123SubdomainWithUnderscore(u.Host) {
+			allErrs = append(allErrs, field.Invalid(endpointFieldPath.Key("host"), u.Host, msg))
+		}
+
+		if u.Path != "" {
+			allErrs = append(allErrs, field.Invalid(endpointFieldPath.Key("path"), u.Path, "endpoint must not have a path component"))
```
zachsmith1, line 88: This validation code forbids paths in endpoints (endpoint must not have a path component), but the controller code attempts to handle paths in the URL parsing

internal/controller/httpproxy_controller.go
```diff
@@ -0,0 +1,452 @@
...
+			Namespace: httpProxy.Namespace,
+			Name:      httpProxy.Name,
+		},
+		Spec: gatewayv1.GatewaySpec{
+			GatewayClassName: r.Config.HTTPProxy.GatewayClassName,
+			Listeners: []gatewayv1.Listener{
+				{
+					Name:     SchemeHTTP,
+					Protocol: gatewayv1.HTTPProtocolType,
+					Port:     gatewayv1.PortNumber(80),
```
zachsmith1, line 275: nit: "dont use magic numbers"

`const (
	SchemeHTTP  = "http"
	SchemeHTTPS = "https"
	
	// HTTP/HTTPS default ports
	DefaultHTTPPort  = 80
	DefaultHTTPSPort = 443
)
...
Port:     gatewayv1.PortNumber(DefaultHTTPPort),
Port:     gatewayv1.PortNumber(DefaultHTTPSPort),
`

internal/controller/httpproxy_controller.go
joshlreese: Do you have some recommended text? Maybe it's best left until we have some documentation to link do?

internal/controller/httpproxy_controller.go
```diff
@@ -0,0 +1,452 @@
...
+				for _, filter := range rule.Filters {
+					if filter.Type == gatewayv1.HTTPRouteFilterURLRewrite {
+						filter.URLRewrite.Hostname = ptr.To(gatewayv1.PreciseHostname(host))
+						hostnameRewriteFound = true
+						break
+					}
+				}
+
+				if !hostnameRewriteFound {
+					rule.Filters = append(backend.Filters, gatewayv1.HTTPRouteFilter{
```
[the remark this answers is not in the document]
joshlreese: Oops, yep. Will add a test to cover this.

internal/controller/httpproxy_controller.go
joshlreese: Fair point - will address those here and in the default backend ports.

internal/controller/httpproxy_controller.go
joshlreese: Thanks for reminding me on this, I was considering adding a limit of 63 characters to these names. K8s otherwise allows for up to 253 characters in the name.

What do you think? I could use a hash, but lean more toward names that are easier for users to understand.

internal/validation/httpproxy_validation.go
joshlreese: I'm not seeing any code remaining for path handling in the controller. We pull out the scheme, port, and host (ip or hostname), but not the path. Am I missing something?

internal/controller/httpproxy_controller.go
zachsmith1: I feel like I've seen the combination of readable plus short hash at the end in a few projects. This way you get readable + unique

internal/controller/httpproxy_controller.go
scotwells: I'm good with updating this once we have documentation to point to. 

I was initially thinking something like this: 

> Underlying Gateway with the name %q already exists and is owned by a different resource. Re-create the HTTP Proxy with a unique name to successfully provision the proxy

But this recommendation isn't super straight forward / easy to understand since they user would need to know how we create gateway resources based on their HTTP Proxy to know how to avoid conflicts.

internal/controller/httpproxy_controller.go
joshlreese: Updated in https://github.com/datum-cloud/network-services-operator/pull/34/commits/b06ca70906eb9ed9d14cb6e5fb3e1bf156ad8e37

internal/controller/httpproxy_controller.go
joshlreese: I've limited HTTPProxy names to 63 characters in https://github.com/datum-cloud/network-services-operator/pull/34/commits/b06ca70906eb9ed9d14cb6e5fb3e1bf156ad8e37

zachsmith1 (approved):
LGTM

scotwells (approved):

PR #65, opened 2025-10-09
feat: wire domain controller with rdap registration

zachsmith1:
This pull request introduces significant improvements to domain registration and nameserver modeling, as well as configuration enhancements for refresh and retry behavior. The changes clarify how authoritative nameservers are represented, refactor the registration data model, and add new configuration options for controlling registration data refresh intervals and backoff strategies.

### Domain and Nameserver Modeling Improvements

* Added `Nameservers` and `Apex` fields to `DomainStatus` to distinguish between apex domains and delegated subdomains, with clear rules for authoritative NS selection. (`api/v1alpha/domain_types.go`, `config/crd/bases/networking.datumapis.com_domains.yaml`) 
* Refactored `Nameserver` to include a list of `NameserverIP` objects, which capture per-address provenance, and removed the direct `Nameservers` field from `Registration`. (`api/v1alpha/domain_types.go`, `api/v1alpha/zz_generated.deepcopy.go`)

### Registration Data Model Updates

* Added `NextRegistrationAttempt` field to `Registration` to track when the next registration refresh should occur. (`api/v1alpha/domain_types.go`, `api/v1alpha/zz_generated.deepcopy.go`, `config/crd/bases/networking.datumapis.com_domains.yaml`)

### Configuration Enhancements

* Introduced `DomainRegistrationConfig` to control RDAP/WHOIS refresh behavior, including refresh interval, retry backoff, jitter, and lookup timeout, with sensible defaults and deep copy support. (`internal/config/config.go`, `internal/config/zz_generated.deepcopy.go`, `internal/config/zz_generated.defaults.go`)

### Test and Default Handling

* Updated tests and object defaulting to ensure new configuration fields are initialized and used correctly in domain verification and registration logic. (`internal/controller/domain_controller_test.go`, `internal/config/zz_generated.defaults.go`)


api/v1alpha/domain_types.go
```diff
@@ -133,6 +136,8 @@ type Registration struct {
 
 	// Abuse / support contacts (registrar/registry)
 	Abuse *AbuseContact `json:"abuse,omitempty"`
+
+	NextRegistrationAttempt metav1.Time `json:"nextRegistrationAttempt,omitempty"`
```
joshlreese: This reads as if it's going to try a domain registration, when it's actually just refreshing registration information. What do you think about `NextRefreshAttempt`, since it's within `Registration`?

internal/controller/domain_controller.go
```diff
@@ -329,6 +353,386 @@ func defaultHTTPGet(ctx context.Context, url string) ([]byte, *http.Response, er
...
+
+	return next
+}
+
+// mapRDAPDomainToRegistration maps a raw RDAP domain into our Registration model.
+func (r *DomainReconciler) mapRDAPDomainToRegistration(d *rdap.Domain) *networkingv1alpha.Registration {
+	reg := &networkingv1alpha.Registration{}
+	if d == nil {
+		return reg
+	}
```
joshlreese: I'm not seeing a scenario where `d` would be nil. If we want to protect against that changing in the future, would it make more sense to return an error instead of an empty registration struct?

PR #81, opened 2025-12-12
feat: registrydata coordination package

zachsmith1:
Rate limits for whois/rdap services are typically pretty low (1QPS) and based on the originating IP. NSO currently fetches this information for each domain and tries to respect rate limiting but unfortunately the parallel nature of the reconcile loop means independent runs dont coordinate their requests for a given whois/rdap provider. This leads to temporary bans on the IPs where that controller runs.

This PR attempts to solve that by implementing a new library where coordination, rate limiting, and retries are handled in a way that leads to safer per provider queries. 

https://github.com/datum-cloud/enhancements/issues/497

### DIFF
--- a/internal/controller/domain_controller.go
+++ b/internal/controller/domain_controller.go
@@ -28,8 +28,8 @@ import (
 
 	networkingv1alpha "go.datum.net/network-services-operator/api/v1alpha"
 	"go.datum.net/network-services-operator/internal/config"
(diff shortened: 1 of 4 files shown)


drewr (dismissed):
Caching is easy right? Right??

Grabbing a couple others just to give it a once-through.

drewr (approved):
Ship it

PR #84, opened 2026-01-06
Introduce Connector CRDs.

joshlreese:
## Summary

Defines CRDs for ConnectorClasses, Connectors, and ConnectorAdvertisements. These have been placed in the `v1alpha1` group and NOT the `v1alpha` group to align with goals on moving away from `v1alpha`.

The definition of capabilities has been slightly adjusted to have a `Disabled` field instead of an `Enabled` field, and to use CamelCase for capability names. Existence of a capability in the spec will imply that it is expected to be enabled, which aligns with best practices on boolean zero values. I’ve also omitted the `leaseRef` field from the status of a Connector, as it will be required to have a matching name anyways. I'll have a separate PR for updating the enhancement to align with these changes.

See: https://github.com/datum-cloud/enhancements/tree/main/enhancements/networking/connectors/initial-proposal

## Other

I have some local changes with a shell script that leverages the [kubectl-validate](https://github.com/kubernetes-sigs/kubectl-validate/tree/main) tool to enable defining tests for CRD validation, including CEL expressions. I'll provide that in a separate PR once I get it to a place that I think is useful. The JSON output of the tool can be used to build specific violation assertions, which my current script doesn't leverage yet.


drewr:
@zachsmith1 What needs to happen before we can merge?

PR #110, opened 2026-02-10
fix: watch Challenges instead of Certificates for ACME HTTP-01 solver

### LINKED ISSUE #109: Certificates not renewing
## Summary

Certificates provisioned for HTTP Proxy / Gateway resources are not automatically renewed because the HTTP Route for the certificate challenge is not provisioned.

## Acceptance Criteria

- Certificates must be renewed automatically when they're due for renewal

scotwells:
The downstream certificate solver controller was watching Certificate resources and exiting early when a certificate had `Ready=True status`. This caused certificate renewals to fail because cert-manager creates new Order and Challenge resources for renewals while the existing certificate is still marked as Ready.

By watching Challenge resources directly, the controller now correctly handles both initial certificate issuance and renewals. When a Challenge is created (for either scenario), the controller:

1. Finds the owning Order via controller reference
2. Gets the Certificate name from the Order's annotation
3. Validates the Certificate uses a configured ClusterIssuer
4. Verifies the Certificate is owned by a Gateway
5. Creates the HTTPRoute and HTTPRouteFilter for the ACME challenge

This approach is also more efficient since reconciliation only occurs when there is actual work to do (a Challenge exists), rather than on every Certificate status change.

---

Closes #109

### DIFF
--- a/internal/controller/gateway_downstream_certificate_solver_controller.go
+++ b/internal/controller/gateway_downstream_certificate_solver_controller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"time"
 
 	envoygatewayv1alpha1 "github.com/envoyproxy/gateway/api/v1alpha1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


zachsmith1:
Nice!!

PR #115, opened 2026-02-24
Add hostname and proxy status conditions to show cert health

mattdjenkinson:
## Summary

Adds visibility into TLS certificate provisioning and status for users of HTTP Proxy resources. Users can see per-hostname certificate state and an aggregate certificate-ready condition on the HTTPProxy.

### API (`api/v1alpha/httpproxy_types.go`)

- **HostnameStatus.Conditions**
  - New condition type: `CertificateReady` (True/False/Unknown).
  - Reasons: `CertificateIssued`, `Pending`, `ProvisioningFailed`, `ChallengeInProgress`.
- **HTTPProxyStatus.Conditions**
  - New condition type: `CertificatesReady` (True when all HTTPS hostnames have ready certificates).
  - Reasons: `AllCertificatesReady`, `CertificatesPending`, `CertificatesFailed`.
- New **Certificates** print column on the HTTPProxy CRD showing `CertificatesReady` status.

### Controller (`internal/controller/httpproxy_controller.go`)

- **buildCertificateStatuses**: Reads cert-manager `Certificate` resources from the downstream cluster for each HTTPS listener and builds per-hostname status with `CertificateReady` set from cert-manager Ready condition and reason.
- **getCertificateReadyConditionReason**: Maps cert-manager Certificate status (e.g. `InvalidCertificate`, `CertificateRequestFailed`) to API reasons.
- **setCertificatesReadyCondition**: Sets or clears the aggregate `CertificatesReady` condition (all ready vs pending vs failed vs unknown when downstream unavailable).
- **reconcileHTTPProxyHostnameStatus** now takes `clusterName`, calls `buildCertificateStatuses`, merges certificate statuses into `HostnameStatuses`, and calls `setCertificatesReadyCondition`.

### RBAC (`config/rbac_downstream/role.yaml`)

- Downstream cluster role updated with **get**, **list**, **watch** on `cert-manager.io` **certificates** so the operator can read certificate status in the downstream cluster.

### Tests (`internal/controller/httpproxy_controller_test.go`)

- **TestBuildCertificateStatuses**: Downstream nil, cert not found, cert ready, cert not ready, cert `InvalidCertificate` (ProvisioningFailed).
- **TestSetCertificatesReadyCondition**: Nil/empty statuses, all ready, any pending, any failed, downstream unavailable with HTTPS listeners.
- **clusterWithClient** test double implementing `cluster.Cluster` for certificate status tests.

## Behaviour

- Certificate status is derived from cert-manager **Certificate** resources in the **downstream** cluster (name: `{gatewayName}-{listenerName}`, namespace: downstream namespace for the HTTPProxy).
- When `DownstreamCluster` is nil, no certificate statuses are built; `CertificatesReady` is set to Unknown when there are HTTPS listeners.
- Per-hostname `CertificateReady` is merged into existing `HostnameStatuses` (with Available and DNSRecordProgrammed) so users see certificate state next to verification and DNS status.

### DIFF
--- a/internal/controller/httpproxy_controller.go
+++ b/internal/controller/httpproxy_controller.go
@@ -1133,7 +1133,14 @@ func getCertificateReadyConditionReason(certificate *unstructured.Unstructured) 
 		switch reason {
 		case "InvalidCertificate", "CertificateRequestFailed":
 			return networkingv1alpha.CertificateReadyReasonProvisioningFailed, message
+		case "DoesNotExist":
+			// cert-manager: "Issuing certificate as Secret does not exist" — normal during first issuance
+			return networkingv1alpha.CertificateReadyReasonPending, "We're provisioning and applying a certificate to this hostname - it may take a few minutes"
 		default:
+			// Translate raw cert-manager message when it's the "Secret does not exist" issuance message
+			if strings.Contains(message, "Issuing certificate as Secret does not exist") {
+				return networkingv1alpha.CertificateReadyReasonPending, "We're provisioning and applying a certificate to this hostname - it may take a few minutes"
+			}
 			return networkingv1alpha.CertificateReadyReasonPending, message
 		}
 	}


scotwells:
@mattdjenkinson its working for me. There was an issue with the deployment process to staging so that change wasn't rolled out until a few mins ago. 

btw @zachsmith1 brought up some good feedback in Slack that users may want to opt-out of the auto delete DNS records behavior. 

> what if the user is migrating away from their httpproxy but still wants that record to be used and pointed at something else? do we still want to auto remove these?

Maybe something to consider or at least tracking in the UI? The UI would just need to issue an update to the DNS records to remove the labels from the DNS Record Set before deleting the HTTP Proxy.

```yaml
  labels:
    app.kubernetes.io/managed-by: networking.datumapis.com
    dns.datumapis.com/managed: "true"
    dns.datumapis.com/source-kind: Gateway
    dns.datumapis.com/source-name: customer-api-7ch71y
    dns.datumapis.com/source-namespace: default
```

scotwells:
@mattdjenkinson after thinking about the deletion lifecycle for HTTP Proxies / DNS Records again, I think we should defer the work for handling this so we don't distract from current progress. 

Retaining the DNS records means the record is going to point to a stale hostname that won't be allocated to the user's proxy anymore meaning they'll need to go in and change it. I don't see that being much different than them having to go re-create the record after deleting the proxy. The number of steps is the same. 

I could see us supporting something like detaching the DNS records from a proxy so the user could change the record to point to a different IP or a different proxy, but that's a bigger chunk of work.

Maybe for now we just warn the user that any DNS records automatically added for custom hostnames on the proxies will automatically be removed?

PR #137, opened 2026-04-13
fix: prevent TPP stale cleanup from deleting connector EPPs

zachsmith1:
Closes datum-cloud/infra#2196

## Root Cause

Confirmed via NSO pod logs on the `bpkrd` pod at `02:14:00Z`:

```
"deleted stale envoypatchpolicy from downstream cluster"
{"controller": "trafficprotectionpolicy", "name": "connector-tunnel-ttqpb"}
```

The `TrafficProtectionPolicy` stale-cleanup loop listed **all** `EnvoyPatchPolicies` in the downstream namespace and deleted any not in its desired set. Whenever a Gateway or HTTPRoute change triggered a TPP reconcile (e.g. on NSO pod failover), it silently deleted `connector-<name>` EPPs that belong to the HTTPProxy controller.

This delete+create cycle then triggered Envoy Gateway's watchable deduplication bug: the recreated EPP has the same name and `generation=1` as the previous one, so EG's in-memory map suppresses the status write → EPP status stays null permanently.

## Fix

Scope the stale-cleanup loop to names with the `"tpp-"` prefix. All EPPs written by the TPP controller are named `tpp-<gateway-name>`; the HTTPProxy controller uses `connector-<name>`. The prefix guard is O(1) per item, requires no label dependencies, and correctly handles deleted gateways whose orphaned EPP would be missed if we only iterated the live gateway list.

## Changes

- Add `tppEnvoyPatchPolicyPrefix = "tpp-"` constant
- Skip EPPs in stale-cleanup loop that don't carry the `"tpp-"` prefix
- Use the constant at the naming site in `getDesiredEnvoyPatchPolicies`
- Regression test: TPP reconcile with no active TPPs deletes stale `tpp-*` EPPs but leaves `connector-*` EPPs untouched

## Previous commits (still included)

The earlier commits on this branch keep the connector EPP alive when a connector goes offline (replacing delete with a `direct_response 503` spec), which eliminates a separate source of delete+create cycles. Both fixes are complementary.

## Test plan

- [x] Deploy with a connector-backed HTTPProxy; trigger a TPP reconcile (e.g. create/delete a Gateway in the same namespace) — confirm `connector-*` EPP is not deleted
- [x] Take the connector offline — confirm EPP spec switches to `direct_response` route and EPP status remains set
- [x] Bring connector back online — confirm EPP spec switches back and status remains `Accepted+Programmed`
- [x] Confirm CONNECT requests return `503 "Tunnel not online"` while connector is offline

### DIFF
--- a/internal/controller/trafficprotectionpolicy_controller.go
+++ b/internal/controller/trafficprotectionpolicy_controller.go
@@ -63,6 +63,13 @@ const (
 	// The stale-cleanup loop uses this prefix to skip EPPs owned by other
 	// controllers (e.g. the HTTPProxy connector controller uses "connector-<name>").
 	tppEnvoyPatchPolicyPrefix = "tpp-"
+
+	// tppManagedLabel is stamped onto every EnvoyPatchPolicy created or updated
+	// by this controller. Once all existing EPPs have been reconciled and carry
+	// this label, the stale-cleanup loop can switch to a label-selector List
+	// instead of the current prefix check, removing the naming-convention
+	// dependency entirely.
+	tppManagedLabel = "networking.datumapis.com/managed-by-tpp-controller"
 )
 
 // certificateReadinessResult contains the result of checking certificate readiness
@@ -176,6 +183,10 @@ func (r *TrafficProtectionPolicyReconciler) Reconcile(ctx context.Context, req N
 		}}
 
 		result, err := controllerutil.CreateOrUpdate(ctx, downstreamStrategy.GetClient(), &policy, func() error {
+			if policy.Labels == nil {
+				policy.Labels = make(map[string]string)
+			}
+			policy.Labels[tppManagedLabel] = "true"
 			policy.Spec = desiredPolicy.Spec
 			return nil
 		})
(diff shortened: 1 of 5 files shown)


internal/controller/trafficprotectionpolicy_controller.go
```diff
@@ -179,21 +185,28 @@ func (r *TrafficProtectionPolicyReconciler) Reconcile(ctx context.Context, req N
...
 	); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to list managed envoypatchpolicies: %w", err)
+		return ctrl.Result{}, fmt.Errorf("failed to list envoypatchpolicies: %w", err)
 	}
 
 	for i := range existingPolicies.Items {
 		existing := &existingPolicies.Items[i]
+		if !strings.HasPrefix(existing.Name, tppEnvoyPatchPolicyPrefix) {
+			continue
+		}
```
scotwells, line 220: Why use a prefix instead of something like a label? Then we could use label selectors for filtering down to the envoy patch policies the controller cares about. Seems like this would potentially cause issues if a user was to prefix their gateway name with `tpp`.

zachsmith1: I dont think we have any concerns on the collision side. A gateway named tpp-foo produces an EPP named tpp-tpp-foo. On labels, agree they're conceptually cleaner and would allow for filtering, but the tpp- naming convention already existed before this PR so existing EPPs in production already carry it. Switching to labels would require a migration or orphan the currently-deployed EPPs

scotwells: Is it worth adding in labels now and keep the prefix so we can leverage labels for filtering?

zachsmith1: ya I think since the prefix naming is already there we can add the labels in and then eventually move away from the prefix only approach. adding that now

scotwells (approved):

PR #138, opened 2026-04-18
feat: add discovery context annotations for project-scoped resources

scotwells:
## Summary

- Adds `discovery.miloapis.com/parent-contexts: Project` annotation to all 13 CRD manifests in `config/crd/bases/`
- Adds `+kubebuilder:metadata:annotations="discovery.miloapis.com/parent-contexts=Project"` kubebuilder marker to all 13 top-level Go resource types

## Why

The Milo API server includes a discovery filter that controls which resources appear in API discovery responses based on the current request context (Platform, Organization, Project, or User). By annotating all networking resources with `parent-contexts: Project`, they will only appear in discovery responses when a client queries within a Project's control plane — keeping the API surface clean and contextually appropriate.

## Affected CRDs

**`networking.datumapis.com` (v1alpha)**
- `Domain`
- `HTTPProxy`
- `Location`
- `Network`
- `NetworkBinding`
- `NetworkContext`
- `NetworkPolicy`
- `Subnet`
- `SubnetClaim`
- `TrafficProtectionPolicy`

**`networking.datumapis.com` (v1alpha1)**
- `Connector`
- `ConnectorAdvertisement`
- `ConnectorClass`

## Notes

The kubebuilder markers were added to the non-List resource types only (i.e., the types with `+kubebuilder:subresource:status`). The CRD YAML files were updated directly to match what `controller-gen` would produce when regenerating from the updated Go markers.


JoseSzycho (approved):

kevwilliams (commented):
Naming and value format match the established convention (discovery.miloapis.com/parent-contexts, compare compute and dns-operator repos), and RBAC/tests are correctly out of scope since this is static CRD/kubebuilder metadata, not runtime behavior.

One scoping concern: Location is cluster-scoped (kubebuilder:resource:scope=Cluster) but is tagged parent-contexts=Project only. Unlike ConnectorClass here, which matches the existing DNSZoneClass precedent of a cluster-scoped "class" companion type getting Project-only visibility, Location is the primary resource representing shared infrastructure (a datacenter/region), not a per-project owned object. The other cluster-scoped precedent in the org, RuntimeClass in compute, uses parent-contexts=Platform,Project rather than Project alone.

Should Location also include Platform in parent-contexts so it stays visible outside a project control plane, or is Project-only intentional here? Worth confirming before merge since this affects where the resource shows up in discovery.

PR #140, opened 2026-04-23
feat: add ActivityPolicies for networking resources

mattdjenkinson:
Add ActivityPolicy manifests covering all 13 networking.datumapis.com kinds (Network, NetworkBinding, NetworkContext, NetworkPolicy, Subnet, SubnetClaim, HTTPProxy, Domain, Location, Connector, ConnectorClass, ConnectorAdvertisement, TrafficProtectionPolicy) and 6 Gateway API kinds (Gateway, HTTPRoute, TCPRoute, UDPRoute, TLSRoute, BackendTLSPolicy). Each policy emits create/delete/update activities from audit logs. Bundled into the existing OCI kustomize artifact via the new config/activity path so infra can deploy them via Flux.

Ref: https://github.com/datum-cloud/enhancements/issues/666


config/milo/activity/policies/httpproxy-policy.yaml
```diff
@@ -0,0 +1,46 @@
...
+spec:
+  resource:
+    apiGroup: networking.datumapis.com
+    kind: HTTPProxy
+
+  auditRules:
+    # HTTPProxy creation with spec.hostnames available - use first hostname as display text
+    - name: create
+      match: "!audit.user.username.startsWith('system:') && audit.verb == 'create' && has(audit.requestObject.spec) && has(audit.requestObject.spec.hostnames) && size(audit.requestObject.spec.hostnames) > 0"
+      summary: "{{ actor }} created HTTP proxy {{ link(string(audit.requestObject.spec.hostnames[0]), audit.objectRef) }}"
```
kevwilliams, line 21: Do we want this to be using the "AI Edge" terminology?

mattdjenkinson: I don't think so because an AI Edge is a collection of resources. @scotwells do you have an opinion on that?

kevwilliams:
My overarching question when looking at this is if we should surface some of these events or if they are internal. I know the System filter lets us filter some out but it still could get quite noisy in some cases, like if a user creates a Connector and the controller creates a ConnectorAdvertisement, that would still look like the human/service did it. Not sure if thats good or not and I think @scotwells would have a good answer. To be specific most of these look totally fine and obviously user-facing, but the ones I am not so sure about are: NetworkBinding, NetworkContext, SubnetClaim, ConnectorClass, ConnectorAdvertisement, BackendTLSPolicy ..

scotwells:
We can drop these resources from the scope of this work since they're not being consumed by end-users today:
- Network
- NetworkContext
- NetworkPolicy
- NetworkBinding
- Subnet
- SubnetClaim
- Location

These are system managed:

- ConnectorClass

All others may be used by consumers today. E.g. the Datum Desktop app will create Connector / Connector Advertisements to expose services running on the user's computer to Datum Cloud. Users may create proxies that proxy requests down to a connector or some other endpoint.

We should also have Olli take a look at these (for some reason can't @ him?)

config/milo/activity/policies/httpproxy-policy.yaml
scotwells: Correct, AI Edge is the product name over a grouping of resources.

scotwells (approved):

PR #161, opened 2026-05-14
fix: tolerate missing Lease discovery on project control planes

### LINKED ISSUE #160: connector/iroh-dns controllers fail to engage on 221 PCPs missing coordination.k8s.io discovery → Connector.Ready frozen indefinitely; 5s retry loop likely driving OOMKill
## Summary

In staging, the `connector` and `iroh-dns` controllers in `network-services-operator` fail to engage for every project control plane whose apiserver does not advertise `coordination.k8s.io/v1` in API discovery. Both controllers register a `Watches(&coordinationv1.Lease{}, …)`, and `mcController.Engage` rejects the whole controller/cluster pair when any watch can't be wired. The result is that for affected clusters:

- `Connector.status.conditions[Ready]` is frozen at the value set the first time the controller successfully engaged (often the creation time, when the Lease hadn't been renewed yet, so `Ready=False(ConnectorNotReady)`).
- `IrohDNSPublished` similarly never updates.
- `metadata.generation` advances but `status.conditions[*].observedGeneration` stays behind.
- HTTP/Gateway/etc. controllers in the same operator binary continue to reconcile that same cluster normally (because they don't watch Lease).

User-visible effect: Datum Connect desktop agents heartbeat correctly every ~15s (Lease `spec.renewTime` is fresh) but the Datum Cloud UI reports the Connector as offline indefinitely.

There is also a secondary `OOMKilled` loop on `controller-manager` (every ~30 min per replica, see "Operator stability" below) which makes the problem worse — every restart re-attempts the failing Engage and re-grows whatever caches are leaking.

## Reproduction

1. In staging, install Datum Connect desktop and create a Connector in a project whose PCP apiserver only advertises `networking.datumapis.com` in `/apis` (i.e. doesn't advertise `coordination.k8s.io`).
2. Verify the agent is patching `spec.renewTime` on the connector's Lease.
3. Observe `Connector.status.conditions[Ready] = False(ConnectorNotReady)` indefinitely. `observedGeneration` lags `metadata.generation`.

## Evidence — affected staging cluster `matt-jenkinson-yz0y92`

Connector `datum-connect-ff98k` (uid `e5b7d11f-945d-4432-8b21-6f66d93f3e5a`, project namespace `default`).

### Connector status (via `kubectl get connector ... -o jsonpath=...`)

```
generation=2  resourceVersion=1138668215
Accepted=True(Accepted)                    obs=1  @ 1970-01-01T00:00:00Z
Ready=False(ConnectorNotReady)             obs=1  @ 2026-03-07T15:20:14Z   ← creation time
IrohDNSPublished=False(DeferredToOwner)    obs=1  @ 2026-05-01T20:11:43Z   ← last operator status write
```

Note `observedGeneration=1` everywhere despite `generation=2`. The controller has never observed the current generation.

`metadata.managedFields` confirms no `manager: manager` (operator) status write has happened since 2026-05-01T20:11:43Z. Datum Desktop (the agent) continues to patch `status.connectionDetails` and renew the Lease.

### Lease (`coordination.k8s.io/v1` `default/datum-connect-ff98k`)

Fetched via `kubectl get --raw …` (kubectl's normal discovery returns `error: the server doesn't have a resource type "leases"` — see "PCP discovery omission" below — but the resource is reachable directly):

```yaml
metadata:
  ownerReferences:
  - apiVersion: networking.datumapis.com/v1alpha1
    kind: Connector
    name: datum-connect-ff98k
    uid: e5b7d11f-945d-4432-8b21-6f66d93f3e5a
    controller: true
    blockOwnerDeletion: true
spec:
  leaseDurationSeconds: 30
  renewTime: 2026-05-14T13:19:58.587566Z   # less than 15s old at fetch
```

Lease is healthy: correct ownerRef, fresh renewTime, valid duration.

### Operator-side logs

The connector/iroh-dns controllers have **never** reconciled `matt-jenkinson-yz0y92` on either of the most recent boots of the leader pod (the boots where they did reconcile httpproxy/gateway for that exact cluster):

```bash
$ kubectl -n datum-system logs <leader-pod> --tail=200000 \
    | grep '"controller":"(connector|iroh-dns)"' | grep matt-jenkinson
$ kubectl -n datum-system logs <leader-pod> --previous --tail=200000 \
    | grep '"controller":"(connector|iroh-dns)"' | grep matt-jenkinson
(empty)
```

For the same boots, the `httpproxy` and `gateway` controllers reconcile the same cluster normally — they don't `Watches(&coordinationv1.Lease{}, …)`, so their Engage isn't rejected.

### Root-cause log line (the bug)

For `matt-jenkinson-yz0y92` specifically, the error repeats **every ~5 seconds in a retry loop** on the leader pod:

```
2026-05-14T13:18:40Z ERROR get informer failed
  {"cluster": "/matt-jenkinson-yz0y92", "source": "kind",
   "error": "no matches for kind \"Lease\" in version \"coordination.k8s.io/v1\""}
2026-05-14T13:18:40Z ERROR cluster-sharding-coordinator failed to engage
  {"cluster": "/matt-jenkinson-yz0y92",
   "error": "failed to watch for cluster \"/matt-jenkinson-yz0y92\":
             no matches for kind \"Lease\" in version \"coordination.k8s.io/v1\""}
2026-05-14T13:18:45Z ERROR get informer failed   { … same … }
2026-05-14T13:18:45Z ERROR cluster-sharding-coordinator failed to engage   { … same … }
2026-05-14T13:18:50Z ERROR get informer failed   { … same … }
2026-05-14T13:18:50Z ERROR cluster-sharding-coordinator failed to engage   { … same … }
…
```

**221 unique project clusters** are in this state on staging (counted via `grep "failed to engage" | grep -oE '"cluster": "/[^"]+"' | sort -u | wc -l` across both boots of all three replicas). The retry-every-5s pattern means each affected cluster generates ~720 error-pair log lines per hour, which also likely contributes to the OOMKill loop below.

A small sample of the 221 affected clusters (alphabetical prefix only):

```
/aaaaaa-d4qxk8
/asdf-6283wa
/asdasd
/e2e-shared-project-1776-{0bjzou,74u9w6,gp25oj,kaghqp,knraaw}
/e2e-shared-project-1777-x69inr
/e2e-test-dns-project-17-jawcmp
/hiyahya-4vrcph
/jacob-test-project-ybdzjo
/jbjjjhji-jm8yi3
/jose-{project-pt1wpv,sirugu}
/matt-jenkinson-yz0y92
/molla-{9rnjfm,otoke-4baody}
/new-project-6x6sz1
/osca-slo-test-r5h1r7
/personal-project-{2119b055,6527428a,759543f8,aeef86da,be933431, …many more}
/tdaly-v20250703-yxt7b6
/test-{delete-n4ccjo,elzw4o,fathom-project-twjb2l,project-{1-fscxij,6z9bj6,quota-8mmaoj,w4t25q},queue-{2-7v9vy1,hupwwe,project-56chve},quota-yu4nc4}
/test{1-clhv7m,2-gbod84,123-xvposa}
/testing-rkuax5
```

## PCP discovery omission

The Datum project-control-plane apiserver only advertises `networking.datumapis.com` in discovery:

```
$ kubectl api-resources --api-group=coordination.k8s.io
(empty)
$ kubectl api-resources --api-group=networking.datumapis.com
connectoradvertisements        networking.datumapis.com/v1alpha1   ConnectorAdvertisement
connectorclasses               networking.datumapis.com/v1alpha1   ConnectorClass
connectors                     networking.datumapis.com/v1alpha1   Connector
…
```

But the underlying Lease resource **is** reachable via direct path:

```
$ kubectl get --raw "/apis/coordination.k8s.io/v1/namespaces/default/leases/datum-connect-ff98k"
{ "kind": "Lease", "apiVersion": "coordination.k8s.io/v1", … }
```

So this is a discovery omission, not a real "Lease isn't there" condition. The Datum Connect desktop uses kube-rs (which doesn't do discovery, just constructs the URL directly) and is consequently able to renew leases without issue.

## Operator stability

Concurrent issue compounding the above: all three replicas of `network-services-operator-controller-manager` are `OOMKilled` every ~30 minutes (Exit 137, `memory: 4Gi` limit). Image `ghcr.io/datum-cloud/network-services-operator:v0.0.0-main-20260512-182158`. Restart counts on `2026-05-14T13:30Z`:

```
network-services-operator-controller-manager-67ff7d4f66-cj8kd   36 restarts in 17h
network-services-operator-controller-manager-67ff7d4f66-j29gb   32 restarts in 17h
network-services-operator-controller-manager-67ff7d4f66-vj4bf   33 restarts in 17h
```

Probably a separate bug (memory growth scaling with number of project clusters / retried engagements). Even if Bug 1 above were fixed, the OOM loop is going to cause stalls.

## Suggested directions

For the engage failure (the primary blocker):

- Make the `Lease` watch optional: catch the discovery error and either continue without it, or schedule a periodic re-attempt without failing the whole `Engage` for the controller. Today, every controller that watches Lease is "all or nothing" per cluster.
- Alternatively, on the PCP apiserver side, advertise `coordination.k8s.io/v1` in `/apis` discovery — since the resource is already reachable, just hidden from clients that do API discovery (kubectl, controller-runtime).
- Either fix would also restore correct reconciles for the iroh-dns controller in these clusters.

For the OOMKill loop — needs a separate investigation; probably worth a pprof heap dump on a healthy-but-near-OOM pod.

## Workaround

None on the agent side. Restarting the operator briefly resurrects reconciles for clusters that *do* engage, but the affected clusters never recover within a pod's lifetime.

mattdjenkinson:
## Summary

`connector` and `iroh-dns` both register `Watches(&coordinationv1.Lease{}, …)` and call `cl.Get(&coordinationv1.Lease{}, …)` inside `Reconcile`. multicluster-runtime rejects the entire controller/cluster pair if any watch can't be wired; controller-runtime's cached client also routes through the cluster's REST mapper. On project control planes that omit `coordination.k8s.io` from API discovery, both paths fail with `no matches for kind "Lease" in version "coordination.k8s.io/v1"` — `Connector.status.conditions[Ready]` freezes, `observedGeneration` lags `generation`, and the operator loops a ~5 s engage-retry storm per affected PCP.

Currently staging-only (~221 affected PCPs). Root cause is upstream in milo: **milo-os/milo#616** — the `DiscoveryContextFilter` feature gate drops built-in groups from project-context aggregated discovery responses because the filter unmarshals legacy `APIGroupList` JSON into an `APIGroupDiscoveryList` struct and emits an empty list. Prod runs with the gate disabled, so it doesn't hit this today.

This PR makes NSO defensive regardless. Even after the milo bug is fixed, this protects against any future downstream apiserver that mis-handles discovery.

## Changes

- **`internal/controller/source/optional.go` (new)** — `OptionalKind` wraps `mcsource.Kind`. On `ForCluster`, it probes the cluster's REST mapper; on `apimeta.IsNoMatchError` it returns `shouldEngage=false` so multicluster-runtime cleanly skips that source for that cluster. Other mapper errors still propagate.
- **`connector_controller.go`, `iroh_dns_controller.go`** — both switch from `.Watches(…Lease…).Complete(r)` to `.Build(r)` plus `c.MultiClusterWatch(OptionalKind(…Lease…))`. The builder's `Watches` always constructs `mcsource.Kind` so we bypass it for Lease only.
- **`ConnectorReconciler.LeaseClient` hook + `ensureConnectorLease`** — Lease access in the reconciler body now goes through `k8s.io/client-go/kubernetes/typed/coordination/v1`. The typed client encodes its own group/version and never asks the REST mapper, so a missing discovery entry no longer breaks `cl.Get(Lease)`. Production constructs the client lazily from `cluster.GetConfig()`; tests inject a `kubefake.NewSimpleClientset` to avoid needing a real REST config.
- **Periodic `RequeueAfter`** — `connectorLeaseReady` now sets `RequeueAfter: leaseDuration + jitter` on every not-ready path, and `IrohDNSReconciler.Reconcile` returns `RequeueAfter: requeueInterval()` for every iroh-eligible Connector. Healthy clusters retain event-driven (sub-second) latency on Lease renew; degraded clusters converge at the lease cadence (≈30 s) instead of never.
- **Unit tests** — `optional_test.go` covers the three branches (no-match degrades, healthy mapper delegates, other errors propagate); the existing `TestConnectorReconcile` cases were updated to inject the fake Lease clientset.

Two minor trade-offs worth noting:

1. The typed Lease client bypasses controller-runtime's cache. Each reconcile makes one extra direct GET for the Lease. At ~200 connectors × ~4 reconciles/min ≈ 13 GET/s — negligible apiserver load.
2. Both reconcilers now requeue at the lease cadence even when the watch is firing normally. Redundant for healthy clusters but cheap, and removes any silent stall path.

## Staging validation (already deployed via [infra#2448](https://github.com/datum-cloud/infra/pull/2448))

Two-commit rollout against the affected 221 PCPs.

After commit `3578c30` (engage fix):

- Zero `failed to engage` / `get informer failed` log pairs across all 3 replicas
- `optional watch unavailable on cluster, degrading to no-op` fires once per cluster at engagement, then silent
- `iroh-dns` reconcile rate 0/s → 0.97/s (was completely stalled; now flowing)
- Controllers engage, but `connector` had a residual 0.1–0.5/s `Reconciler error` rate from `cl.Get(Lease)` still hitting the RESTMapper

After commit `2bedb7a` (typed-client fix):

- `connector` `Reconciler error` rate ~0/s (single 0.003/s blip on one pod = noise)
- Per-pod memory 0.4–0.5 GiB (was OOMKilling at 4 GiB every ~30 min pre-fix)
- 58 min post-rollout, zero restarts on the new ReplicaSet
- `matt-jenkinson-yz0y92` (one of the affected projects from the issue repro) confirmed reconciling cleanly in logs at 15:23:57Z+

## Test plan

- `go build ./...`, `go vet ./...`, `go test ./...` all clean locally
- New unit tests cover the `OptionalKind` branches and the typed Lease client injection
- Staging deploy validated above

Fixes #160. Related: milo-os/milo#616 (upstream root cause).

### DIFF
--- a/internal/controller/connector_controller.go
+++ b/internal/controller/connector_controller.go
@@ -14,6 +14,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	coordinationv1client "k8s.io/client-go/kubernetes/typed/coordination/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/cluster"
@@ -34,6 +35,26 @@ import (
 type ConnectorReconciler struct {
 	mgr    mcmanager.Manager
 	Config config.NetworkServicesOperator
+
+	// LeaseClient builds a typed Lease client for a given project cluster.
+	// Production leaves this nil; the reconciler falls back to
+	// coordinationv1client.NewForConfig(cl.GetConfig()). The typed client
+	// encodes its own group/version, so it does not depend on the cluster's
+	// REST mapper — that's the whole point of routing Lease access through
+	// it rather than the controller-runtime client, which would fail with
+	// `no matches for kind "Lease"` against project control planes that
+	// omit coordination.k8s.io from discovery (see
+	// network-services-operator#160). Tests inject a fake clientset's
+	// CoordinationV1() so they don't need a real REST config.
+	LeaseClient func(cluster.Cluster) (coordinationv1client.LeasesGetter, error)
+}
+
+// leases returns the typed Lease client for the supplied cluster.
+func (r *ConnectorReconciler) leases(cl cluster.Cluster) (coordinationv1client.LeasesGetter, error) {
+	if r.LeaseClient != nil {
+		return r.LeaseClient(cl)
+	}
+	return coordinationv1client.NewForConfig(cl.GetConfig())
 }
 
 // +kubebuilder:rbac:groups=networking.datumapis.com,resources=connectors,verbs=get;list;watch;create;update;patch;delete
@@ -120,31 +141,21 @@ func (r *ConnectorReconciler) Reconcile(ctx context.Context, req mcreconcile.Req
 		return ctrl.Result{}, nil
 	}
 
+	leases, err := r.leases(cl)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("build lease client: %w", err)
+	}
+
 	leaseDurationSeconds := r.connectorLeaseDurationSeconds()
 	if connector.Status.LeaseRef == nil || connector.Status.LeaseRef.Name == "" {
-		lease := &coordinationv1.Lease{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      connector.Name,
-				Namespace: connector.Namespace,
-			},
-		}
-
-		if _, err := controllerutil.CreateOrUpdate(ctx, cl.GetClient(), lease, func() error {
-			if err := controllerutil.SetControllerReference(&connector, lease, cl.GetScheme()); err != nil {
-				return err
-			}
-			if lease.Spec.LeaseDurationSeconds == nil || *lease.Spec.LeaseDurationSeconds == 0 {
-				lease.Spec.LeaseDurationSeconds = &leaseDurationSeconds
-			}
-			return nil
-		}); err != nil {
+		lease, err := r.ensureConnectorLease(ctx, cl, leases, &connector, leaseDurationSeconds)
+		if err != nil {
 			return ctrl.Result{}, err
 		}
-
 		connector.Status.LeaseRef = &corev1.LocalObjectReference{Name: lease.Name}
 	}
 
-	leaseStatus, err := r.connectorLeaseReady(ctx, cl.GetClient(), &connector)
+	leaseStatus, err := r.connectorLeaseReady(ctx, leases, &connector)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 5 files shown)


scotwells:
@mattdjenkinson interesting side-effect of the discovery stuff! Let's add leases back to projects from a discovery perspective to fix this.

scotwells:
@mattdjenkinson lets close this and fix on the milo side.

PR #156, opened 2026-05-11
docs: design brief for service catalog registration

kevwilliams:
Closes #155 (design/sign-off phase only — no implementation).

## Summary

- Adds `docs/enhancements/service-catalog-registration.md`, a design brief for registering Network Services in the Datum Cloud service catalog.
- Covers all three parts of issue #155: Service registration, metering configuration, and usage pipeline integration investigation.
- Splits work into Phase 1 (pure YAML catalog declarations, no Go changes) and Phase 2 (usage emission from the edge proxy).
- Incorporates Scot's review feedback and resolves all four Phase 1 open decisions. Phase 1 is unblocked.

## What the brief covers

**Catalog declarations (Phase 1, unblocked)**
- `Service` resource (`networking.datumapis.com`) making Network Services discoverable in the catalog.
- `ServiceConfiguration` resource (`services.miloapis.com/v1alpha1`) declaring `HTTPRoute` as the billable Kind and the four meters inline. A fan-out controller produces the billing `MonitoredResourceType` and meter resources — we do not author them directly. This matches the production shape already used by the `compute` service.
- Four meters under the `networking.datumapis.com/gateway/*` family: request count, egress bytes, ingress bytes, connection seconds.
- Dimensions: `region`, `gateway`, `gateway_class` (all optional).
- Kustomize bundle at `config/services/networking.datumapis.com/`, matching `datum-cloud/datum/config/services/<service-domain>/`.

**Why HTTPRoute (not HTTPProxy or Gateway)**
- An HTTP request is what is actually being processed and billed.
- `Gateway` is the underlying implementation; advanced consumers may interact with it directly, but it is not the natural per-request billing unit.
- Modelling at the route layer leaves room for future `TCPRoute`/`UDPRoute` monitored resources with distinct meter families if TCP/UDP traffic is priced differently — which is the dominant industry pattern (see provider survey in the brief: AWS ALB/NLB, GCP HTTPS LB / Network LB, Cloudflare Workers / Spectrum, Azure App Gateway / Standard LB, Fastly).

**Usage pipeline integration (Phase 2 investigation)**
- Three candidate approaches evaluated: Vector access log scraping (recommended), Prometheus counter polling, OTel OTLP sink (rejected per billing pipeline design).
- Connection-seconds for WebSocket / long-lived connections handled via a controller-side connection-lifecycle emitter.
[...]

Issue #155, opened 2026-05-08
feat: register Network Services in the service catalog


scotwells:
## Summary

Network Services needs to be registered in the Datum Cloud service catalog so it can be discovered, enabled, and billed through the platform. This covers three pieces of work:

1. **Service registration** — create the `Service` resource that makes Network Services visible in the catalog with a display name, description, and published phase.

2. **Metering configuration** — create the `ServiceConfiguration` resource that defines what gets measured and billed. For an edge proxy, the most important signals to capture are:

   - **Request count** — number of requests proxied, the primary unit of consumption
   - **Bytes transferred** — total response bytes sent to clients (egress)
   - **Bytes received** — total request bytes received from clients (ingress)
   - **Active connection seconds** — time connections are held open, useful for persistent/long-lived connections

3. **Usage pipeline integration** — investigate how to extract the above signals from the proxy and feed them into the billing pipeline. This includes:

   - How to collect usage data from the proxy (e.g. access logs, stats endpoint, gRPC metrics sink)
   - How to transform and emit that data as usage events in the CloudEvents format expected by the billing pipeline
   - Any buffering, batching, or reliability considerations for getting events from the data plane to the pipeline

Both the `Service` and `ServiceConfiguration` resources should be packaged as a Kustomize component (following the pattern established in the compute service) so they can be applied consistently across environments.

## Acceptance criteria

- [ ] Service appears in the platform-wide service catalog in Published phase
- [ ] Metering configuration covers request count, egress bytes, ingress bytes, and connection seconds
- [ ] Resources are packaged as a `service-catalog` Kustomize component under `config/components/`
- [ ] Investigation complete with a clear approach for collecting and emitting usage events from the proxy to the billing pipeline

PR #165, opened 2026-05-22
feat: register Network Services in the service catalog (Phase 1)

kevwilliams:
Implements Phase 1 of the service-catalog-registration design brief (PR #156, issue #155): pure YAML, no Go changes.

## What this PR adds

A Kustomize bundle under `config/services/` that registers Network Services in the platform service catalog and declares its monitored resource and meters:

- **`Service`** (`services.miloapis.com/v1alpha1`)
  - `serviceName: networking.datumapis.com`
  - `phase: Draft`
  - `owner.producerProjectRef.name: datum-cloud`
- **`ServiceConfiguration`** (`services.miloapis.com/v1alpha1`)
  - Monitored resource: `HTTPRoute` (`gateway.networking.k8s.io`)
  - Six optional dimensions: `region`, `gateway`, `gateway_namespace`, `gateway_class`, `httproute_name`, `httproute_namespace`
  - Four meters inline under `networking.datumapis.com/gateway/*`: `requests`, `egress-bytes`, `ingress-bytes`, `connection-seconds`

## Bundle layout

```
config/services/
  kustomization.yaml
  networking.datumapis.com/
    kustomization.yaml      # kind: Component
    service.yaml
    serviceconfiguration.yaml
```

Matches the per-service-domain layout used by `compute` in `datum-cloud/datum`. The inner `kustomization.yaml` is a Kustomize `Component`, mirroring the compute reference.

## What this PR does NOT add

- No `MonitoredResourceType` or `MeterDefinition` resources directly — those are produced by the catalog's fan-out controller from `ServiceConfiguration.spec`, so no `config/billing/` bundle is needed.
- No Go code changes — Phase 1 is YAML-only.
- No Flux wiring — that is a separate step in `datum-cloud/infra` handled by the platform/infra team after sign-off.

Phase 2 (usage emission from the edge proxy) remains blocked on OD-5 through OD-8 in the design brief.

## Test plan

- [x] `kubectl kustomize config/services/` renders both resources without error
- [ ] Reviewer confirms `Service` and `ServiceConfiguration` shapes match the catalog CRDs in the deployment target
- [ ] After merge: platform/infra wires `config/services/` into Flux; confirm the fan-out controller produces the expected `MonitoredResourceType` and `MeterDefinition` resources in the billing namespace
- [ ] Confirm `Service` appears in the catalog in `Draft` phase

Refs #155. Design brief: PR #156.

### DIFF
--- a/config/services/kustomization.yaml
+++ b/config/services/kustomization.yaml
@@ -1,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+components:
+  - networking.datumapis.com
--- a/config/services/networking.datumapis.com/kustomization.yaml
+++ b/config/services/networking.datumapis.com/kustomization.yaml
@@ -1,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+
+resources:
+  - service.yaml
+  - serviceconfiguration.yaml
--- a/config/services/networking.datumapis.com/service.yaml
+++ b/config/services/networking.datumapis.com/service.yaml
@@ -1,0 +1,19 @@
+apiVersion: services.miloapis.com/v1alpha1
+kind: Service
+metadata:
+  name: networking-datumapis-com
+  labels:
+    app.kubernetes.io/name: network-services-operator
+    app.kubernetes.io/managed-by: kustomize
+spec:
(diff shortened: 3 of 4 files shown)


kevwilliams:
I don't think we need this anymore, we can always reopen if we do but the design has changed a bit so its fair for us to redo this as well it was low effort.

Issue #167, opened 2026-05-22
Tunnel routing not live until next connector heartbeat fires after Programmed=True


drewr:
## This is part of a four-issue tunnel creation story

A newly created tunnel takes up to ~14 minutes before it reliably routes traffic. There are two distinct delays, each with an operator-side and a client-side component:

- **Delay 1 (~3-4 min): creation → toggle turns green**
  - [network-services-operator#166](https://github.com/datum-cloud/network-services-operator/issues/166) — NSO reconcile loop sleeps after 409 conflict burst
- **Delay 2 (~0-10 min): toggle green → traffic flows**
  - [network-services-operator#167](https://github.com/datum-cloud/network-services-operator/issues/167) — NSO waits for connector heartbeat instead of pushing addressing proactively ← **this issue**
  - [app#159](https://github.com/datum-cloud/app/issues/159) — heartbeat interval can reach ~10 min
- **UX consequence**
  - [app#160](https://github.com/datum-cloud/app/issues/160) — green toggle shown before tunnel is usable

---

## Summary

After the NSO sets `Programmed=True` on an HTTPProxy, traffic routing through the iroh-gateway is **not confirmed live** until the connector (Datum Desktop) next posts a status heartbeat. The operator reconciles all HTTPProxies backed by a connector in response to connector status updates, not proactively when marking a proxy programmed.

## Observed

For `tunnel-xnhnb` (project `drewr-y4nd1b`), 2026-05-22:

| Event | Timestamp | Delta from Programmed |
|-------|-----------|----------------------|
| `Programmed=True` set on HTTPProxy | 19:37:37Z | — |
| `ConnectorMetadataProgrammed=True` set | 19:37:37Z | — |
| Datum Desktop patches connector `datum-connect-jttwh` status | 19:48:28Z | **+10m51s** |
| NSO sweeps all HTTPProxies for that connector | 19:48:28Z | +10m51s |

The UI toggle turns green at 19:37:37Z. The tunnel does not reliably route traffic until ~19:48:28Z.

## Root Cause

The iroh-gateway resolves `connector.local` using the connector's Iroh node key (public key + relay addresses). The NSO pushes current connector addressing to downstream routes in response to connector status updates, not proactively on HTTPProxy reconcile completion. After `Programmed=True` is set, there is no operator activity until the Datum Desktop client posts a heartbeat.

[...]

PR #169, opened 2026-05-22
fix: requeue quickly after 409 conflict in HTTPProxy reconcile

### LINKED ISSUE #166: HTTPProxy reconcile backs off 3-4 min after 409 conflict burst at tunnel creation
## This is part of a four-issue tunnel creation story

A newly created tunnel takes up to ~14 minutes before it reliably routes traffic. There are two distinct delays, each with an operator-side and a client-side component:

- **Delay 1 (~3-4 min): creation → toggle turns green**
  - [network-services-operator#166](https://github.com/datum-cloud/network-services-operator/issues/166) — NSO reconcile loop sleeps after 409 conflict burst ← **this issue**
  - [network-services-operator#167](https://github.com/datum-cloud/network-services-operator/issues/167) — NSO waits for connector heartbeat instead of pushing addressing proactively
- **Delay 2 (~0-10 min): toggle green → traffic flows**
  - [network-services-operator#167](https://github.com/datum-cloud/network-services-operator/issues/167) — NSO waits for connector heartbeat instead of pushing addressing proactively
  - [app#159](https://github.com/datum-cloud/app/issues/159) — heartbeat interval can reach ~10 min
- **UX consequence**
  - [app#160](https://github.com/datum-cloud/app/issues/160) — green toggle shown before tunnel is usable

---

## Summary

When a new tunnel (HTTPProxy) is created, the NSO fires a burst of writes to the child Gateway and HTTPRoute. Multiple goroutines race and produce several `409 Conflict` errors, after which the controller waits for its next **periodic requeue tick (~3-4 min)** rather than re-queuing immediately on conflict resolution.

## Observed

Traced via audit log on tunnel `tunnel-xnhnb` (project `drewr-y4nd1b`), 2026-05-22:

| Tunnel | Created | Accepted | Programmed | Gap |
|--------|---------|----------|------------|-----|
| tunnel-xnhnb | 19:33:47Z | 19:33:50Z (+3s) | 19:37:37Z | **3m47s** |
| tunnel-79nsr | 18:19:05Z | 18:19:08Z (+3s) | 18:22:09Z | 3m01s |
| tunnel-hz2v8 | 18:21:47Z | 18:21:49Z (+2s) | 18:24:06Z | 2m17s |

The operator writes to `default/gateways/tunnel-xnhnb` and `default/httproutes/tunnel-xnhnb` from `19:33:47Z–19:33:52Z` (several 409s), then goes completely silent until `19:37:37Z` — exactly the periodic requeue window — when it sets `Programmed=True`.

## Root Cause

After the 409 conflict burst resolves, no event-driven trigger wakes the reconcile loop early. The controller relies on its periodic requeue interval, which is empirically 2-4 minutes.

## Expected

After a 409 conflict, the controller should requeue with a short explicit duration (e.g. 5s) so the next reconcile attempt fires quickly once the write conflict resolves. The total creation → `Programmed=True` time should be well under 30s.

## Impact

Every tunnel creation imposes a mandatory ~3-4 minute wait before the UI toggle turns green and the user can interact with the tunnel.

drewr:
## Summary

Fixes #166.

When a new tunnel (HTTPProxy) is created, concurrent reconciles race to write the child Gateway, HTTPRoute, and EndpointSlice. The resulting 409 Conflict errors were passed back to controller-runtime as plain errors, which applied **exponential backoff** — after ~15 conflicts in the initial burst, the backoff reached 3-4 minutes. The controller went silent until the next periodic tick.

This was the root cause of the ~3m47s delay between tunnel creation and `Programmed=True` being set on the HTTPProxy, which is when the UI toggle turns green.

### Before

```
19:33:47Z  HTTPProxy created
19:33:47–52Z  burst of 409 Conflict writes to Gateway + HTTPRoute
19:33:52Z  controller goes silent (exponential backoff accumulated)
19:37:37Z  next tick: Programmed=True set (~3m47s total)
```

### After

Each 409 Conflict returns `ctrl.Result{RequeueAfter: 1s}` instead of entering the exponential-backoff queue. The conflict resolves within one retry cycle; `Programmed=True` should be set within seconds of the child resources being accepted.

## Changes

- **Named the `ctrl.Result` return** from `_ ctrl.Result` to `result ctrl.Result` so the deferred status-update block can set `result.RequeueAfter` on conflict without joining an error that re-enters the backoff queue.
- **Added `apierrors.IsConflict`** guard in the Gateway, HTTPRoute, and EndpointSlice `CreateOrUpdate` error paths — returns `ctrl.Result{RequeueAfter: retryAfterConflict}` (1s), matching the pattern in `gateway_dns_controller.go` and `result.go`.
- **Renamed `result` locals** (the `controllerutil.OperationResult` string type) to `opResult` to avoid shadowing the named return.
- **Added `TestHTTPProxyReconcileConflictRequeue`** covering all three write paths (Gateway update, HTTPRoute update, status update) with injected 409s.

## Testing

```
go test ./internal/controller/ -run TestHTTPProxyReconcileConflictRequeue -v
```

All three subtests pass; full controller suite clean.

## Related

- Closes #166
- Part of the four-issue tunnel creation story: [#166](https://github.com/datum-cloud/network-services-operator/issues/166) → [#167](https://github.com/datum-cloud/network-services-operator/issues/167) → [app#159](https://github.com/datum-cloud/app/issues/159) + [app#160](https://github.com/datum-cloud/app/issues/160)

### DIFF
--- a/internal/controller/httpproxy_controller.go
+++ b/internal/controller/httpproxy_controller.go
@@ -88,7 +88,7 @@ const (
 // +kubebuilder:rbac:groups=gateway.envoyproxy.io,resources=httproutefilters,verbs=get;list;watch;create;update;patch;delete
 // HTTPProxy controller reads cert-manager Certificate resources in the downstream cluster for status; ensure downstream role has cert-manager.io/certificates get;list;watch.
 
-func (r *HTTPProxyReconciler) Reconcile(ctx context.Context, req mcreconcile.Request) (_ ctrl.Result, err error) {
+func (r *HTTPProxyReconciler) Reconcile(ctx context.Context, req mcreconcile.Request) (result ctrl.Result, err error) {
 	logger := log.FromContext(ctx, "cluster", req.ClusterName)
 	ctx = log.IntoContext(ctx, logger)
 
@@ -168,9 +168,18 @@ func (r *HTTPProxyReconciler) Reconcile(ctx context.Context, req mcreconcile.Req
 		if !equality.Semantic.DeepEqual(httpProxy.Status, httpProxyCopy.Status) {
 			httpProxy.Status = httpProxyCopy.Status
 			if statusErr := cl.GetClient().Status().Update(ctx, &httpProxy); statusErr != nil {
-				err = errors.Join(err, fmt.Errorf("failed updating httpproxy status: %w", statusErr))
+				if apierrors.IsConflict(statusErr) {
+					// Optimistic-locking conflict on the status write: requeue
+					// quickly rather than entering the exponential-backoff queue.
+					// This prevents multi-minute stalls when concurrent reconciles
+					// race to update the same HTTPProxy status on creation.
+					result.RequeueAfter = retryAfterConflict
+				} else {
+					err = errors.Join(err, fmt.Errorf("failed updating httpproxy status: %w", statusErr))
+				}
+			} else {
+				logger.Info("httpproxy status updated")
 			}
-			logger.Info("httpproxy status updated")
 		}
 	}()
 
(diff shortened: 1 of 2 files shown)


drewr:
@savme Seeing if this passes the smell test on some issues I've been having lately with creating tunnels. Thanks!

scotwells:
This feels heavy handed. We need to understand why conflicts are happening before we just throw requeues at the problem. Seems like we should be using server side apply or better conflict resolution.

drewr:
Copy that. Incoming.

PR #177, opened 2026-06-09
docs: add WASM filter emitter as the recommended approach for telemetry and billing attribution

JoseSzycho:
This PR adds a new alternative to the metering design document.

Related to: 
- https://github.com/datum-cloud/network-services-operator/pull/156

### DIFF
--- a/docs/enhancements/service-catalog-registration.md
+++ b/docs/enhancements/service-catalog-registration.md
@@ -578,11 +578,15 @@ are needed.
 
 ### Recommendation
 
-**Option A** (access log scraping via Vector) is the most consistent with the
-billing pipeline architecture and is recommended as the primary collection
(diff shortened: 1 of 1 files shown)


kevwilliams:
Very nice approach

scotwells (changes requested):
When we get to a scale that aggregation is necessary, we'll add aggregation as a platform capability in the pipeline instead of something that's envoy specific. There's a lot this proposal isn't considering around the durability needs of the pipeline when something goes wrong. 

At the current scale, per-request events without aggregation will be just fine.

JoseSzycho:
@scotwells thanks for the review. I just updated the doc to leave the recommendation as it was before (Option A).

JoseSzycho:
@kevwilliams @scotwells just updated the doc that Option A with out aggregation would be fine as an initial implementation.

scotwells (approved):

Issue #174, opened 2026-06-07
IrohDNSPublished=True flips to DeferredToOwner mid-session on an actively-serving Connector


drewr:
## Summary

I had two cases with different semantics where an otherwise healthy tunnel stopped functioning after hours. After adding more logging and diagnosing other issues, the second of these cases revealed that there was a collision in resources _across two different Datum Cloud accounts_.

Much of the following is Opus 4.7 context that needed to save somewhere. But the two biggest needs I have from this issue are:
- Potentially merge #160, as it has a fix for part of the issue
- A PR which more fully fixes the issue (thanks @savme for #178!)

## Observed behavior

On a Connector with `IrohDNSPublished=True; Reason=Owner` that has been serving traffic, the condition periodically flips to:

```
Type:    IrohDNSPublished
Status:  False
Reason:  DeferredToOwner
Message: iroh DNS record is owned by Connector /<project>/default/<other-connector>
         (uid <other-uid>)
```

The transition occurs on a server-side reconcile. No client action triggers it (no Create/Update/Delete on the affected Connector or its dependencies). After the flip, traffic to the corresponding tunnel returns 5xx at the edge until ownership is reclaimed.

Two cases of this behavior have been observed in the field. They share the same user-visible pattern but the cited "owner" Connector is in different states in each.

---

### Case 1 — cited owner does not exist in the same project

- Active Connector: `datum-connect-ltwr5`, project `drewr-y4nd1b`, UID `e43a8dd4-b10e-42de-b9ff-032788491a24`.
- Cited owner: `datum-connect-jttwh`, same project `drewr-y4nd1b`, UID `226a90b6-3cad-4242-9eff-c2c71a335545`.
- `datumctl get connector datum-connect-jttwh --project drewr-y4nd1b -n default` returns NotFound (not a permission error — same auth context that successfully resolves `ltwr5`).
- `IrohDNSPublished` on `ltwr5` was `True; Owner` from `2026-06-07T14:44:39Z`, flipped to `False; DeferredToOwner` at `2026-06-07T18:17:58Z` — ~3.5 hours of active operation before the flip.
- Live HTTPProxy tied to the affected Connector (`tunnel-kl6gj`, hostname `talk-known-phk4r.datumproxy.net`) returned HTTP 503 from the moment the flip was visible, until recovery.

### Case 2 — cited owner is a live Connector in a different user's project; ownership oscillates

[...]

PR #178, opened 2026-06-09
fix: competing owner lease takeover

savme:
Related to https://github.com/datum-cloud/network-services-operator/issues/174

## Problem

When multiple connectors share the same iroh public key, the iroh-dns controller uses a first-writer-wins claim on the downstream `DNSRecordSet`. The claim was tracked by UID label only - there was no way to tell whether the claimant was still alive. A connector in a foreign project with a dead or deleted agent could hold the claim indefinitely, blocking an active connector from publishing its DNS record and causing 5xx at the edge.

Two production issues confirmed this: one where the cited owner had been deleted and the active connector never recovered, and one where a cross-account connector with an expired lease displaced an actively-serving session.

## Fix

Ownership arbitration needs a liveness signal that any reconciler can check regardless of which project cluster the claimant lives in. The downstream DNS cluster is the one place all reconcilers can reach by design, so liveness state now lives there: the claim holder writes a `Lease` next to its `DNSRecordSet` and renews it on every heartbeat. A competitor checks that Lease before deferring - if it's absent or expired, the competitor takes over.

### DIFF
--- a/config/rbac_downstream/role.yaml
+++ b/config/rbac_downstream/role.yaml
@@ -182,3 +182,13 @@ rules:
   - envoypatchpolicies/status
   verbs:
   - get
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - create
+  - delete
+  - get
+  - patch
+  - update
--- a/internal/controller/iroh_dns_controller.go
+++ b/internal/controller/iroh_dns_controller.go
@@ -69,7 +69,17 @@ const (
 	claimStatusCreate   claimStatus = iota // no existing record
 	claimStatusOwner                       // we already hold the claim
 	claimStatusTakeOver                    // downstream claim, owner dead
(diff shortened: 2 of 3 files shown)


drewr:
Thanks @savme! For context to reviewers, this is (hopefully) going to address the remaining stability issue with tunnels. Check the linked issue for more detail, but essentially a tunnel can be killed by a zombie connector _in a totally different user account_.

scotwells:
I wonder if we have a gap in the design? Seems odd that the same public key can be used across multiple connectors?

drewr:
The way you worded that makes me realize that a local bug I'm chasing might be contributing here. Let's not merge this until I can verify.

drewr:
I'm going to chalk this one up to a bug where I could have been using the same key across connectors during development for the connect plugin. I haven't hit it since then. Appreciate the work @savme and the catch @scotwells!

PR #181, opened 2026-06-10
docs: refactor service registration brief to HTTP traffic metering

JoseSzycho:
> [!WARNING]
> In `pause` till https://github.com/datum-cloud/network-services-operator/pull/187 is on place
> https://github.com/datum-cloud/network-services-operator/pull/181#discussion_r3428905145

This PR refactors and expands the initial service catalog registration design brief into a structured, standard enhancement proposal focusing on **HTTP Traffic Metering** for Network Services.

It addresses and resolves all open questions (OD-1 through OD-8) carried over from the previous design brief.

Related to:
- https://github.com/datum-cloud/network-services-operator/pull/156#issuecomment-4665223603

Working implementation using OTLP Transport:
- https://github.com/datum-cloud/network-services-operator/pull/188


docs/enhancements/http-traffic-metering.md
```diff
@@ -0,0 +1,391 @@
...
+
+## Motivation
+
+Network Services is a core utility that incurs direct infrastructure costs. Capturing consumption signals is necessary for platform billing and cost-attribution. 
+
+Because `MeterDefinition` fields (such as `meterName` and `measurement.unit`) are immutable once published, establishing correct definitions in the `Draft`/`Provisional` phase is critical. Doing so avoids costly SDK upgrades, meter deprecation cycles, and data migrations.
+
+### Goals
+
+- Define a standard `Service` and `ServiceConfiguration` to register Network Services under the service domain `networking.datumapis.com`.
```
scotwells: This is more of an implementation detail. I'd keep goals high level / consumer focused.

docs/enhancements/http-traffic-metering.md
```diff
@@ -0,0 +1,391 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Network Services operates an Envoy Gateway-based edge proxy that routes HTTP traffic, terminates TLS, and enforces WAF and rate-limit policies on behalf of platform customers. Today, it lacks a billing presence: there is no registration in the service catalog, no meter definitions, and no integration with the durable usage pipeline.
+
+This enhancement defines the architecture, data structures, and roadmap to bring HTTP traffic metering and catalog registration to Network Services. The work is split into two phases:
+- **Phase 1 (Catalog & Metadata):** Declare a `Service` and a companion `ServiceConfiguration` resource (`services.miloapis.com/v1alpha1`) carrying the monitored-resource and meter declarations inline. This is a YAML-only delivery packaged in the `config/services/` bundle.
+- **Phase 2 (Emission & Integration):** Configure Envoy Gateway proxy logging and deploy a custom Vector Agent to scrape access logs, parse billing signals into CloudEvents, and forward them to the local `billing-usage-collector-vector` DaemonSet.
```
scotwells: I don't think it's important to highlight the phases in the enhancement. How the work breaks down and is sequenced is better for a project management tool (e.g. GitHub issues), not an enhancement document.

docs/enhancements/http-metering/http-traffic-metering.md
```diff
@@ -0,0 +1,391 @@
...
+
+This enhancement defines the architecture, data structures, and roadmap to bring HTTP traffic metering and catalog registration to Network Services. The work is split into two phases:
+- **Phase 1 (Catalog & Metadata):** Declare a `Service` and a companion `ServiceConfiguration` resource (`services.miloapis.com/v1alpha1`) carrying the monitored-resource and meter declarations inline. This is a YAML-only delivery packaged in the `config/services/` bundle.
+- **Phase 2 (Emission & Integration):** Configure Envoy Gateway proxy logging and deploy a custom Vector Agent to scrape access logs, parse billing signals into CloudEvents, and forward them to the local `billing-usage-collector-vector` DaemonSet.
+
+## Motivation
+
+Network Services is a core utility that incurs direct infrastructure costs. Capturing consumption signals is necessary for platform billing and cost-attribution. 
+
+Because `MeterDefinition` fields (such as `meterName` and `measurement.unit`) are immutable once published, establishing correct definitions in the `Draft`/`Provisional` phase is critical. Doing so avoids costly SDK upgrades, meter deprecation cycles, and data migrations.
```
scotwells, line 46: This feels like internal / implementation detail. The motivation should be more product / consumer focused.

docs/enhancements/http-traffic-metering.md
```diff
@@ -0,0 +1,391 @@
...
+- Choose a scalable, low-risk telemetry collection approach that bridges Envoy Gateway's telemetry and the Billing Ingestion Gateway.
+- Map the end-to-end data flow with sequence and architecture diagrams.
+
+### Non-Goals
+
+- Defining rate cards, pricing models, tiers, or billing/invoice generation logic.
+- Altering the `MeterDefinition` schema or billing pipeline contract.
+- Implementation of the core Billing SDK (owned by the Billing Team).
+- Shared-infrastructure cost attribution or cross-project billing logic.
+- **Deploying or modifying the Billing System or `billing-usage-collector-vector` DaemonSet.** These components are pre-existing, shared platform infrastructure. Our work is limited to deploying a custom Vector Agent (Log Parser) to parse logs and forward them to this existing collector.
```
scotwells: I don't think this should be a non-goal, it's more of an implementation detail.

docs/enhancements/http-traffic-metering.md
```diff
@@ -0,0 +1,391 @@
...
+
+## Proposal
+
+We propose to register the service and implement HTTP traffic metering via access log scraping. 
+
+- The **Monitored Resource** is the Kubernetes Gateway API `HTTPRoute` resource, representing the customer-facing HTTP endpoint.
+- **Phase 1** registers the service with the service catalog via declarative YAML configurations. The service catalog fan-out controller automatically creates `MonitoredResourceType` and `MeterDefinition` resources in the billing namespace.
+- **Phase 2** instruments the Envoy Gateway instances to write structured JSON access logs to stdout. A node-level Vector Agent (Log Parser) tails these logs, parses and maps the raw logs into CloudEvents, and forwards them locally via HTTP to the `billing-usage-collector-vector` DaemonSet. The billing collector then handles local disk buffering and reliably forwards them to the Billing System.
```
scotwells: We should measure the resource impact this would have on the worker node. It may make sense to have the existing vector agent handle collection from envoy and transforming the results into billable usage events.

docs/enhancements/http-metering/http-traffic-metering.md
```diff
@@ -0,0 +1,391 @@
...
+- `region`: Deployment region.
+- `gateway`: Name of the parent Gateway.
+- `gateway_namespace`: Namespace of the parent Gateway.
+- `gateway_class`: Underlying GatewayClass (for pricing class differentiation).
+- `httproute_name`: The `HTTPRoute` resource name.
+- `httproute_namespace`: The `HTTPRoute` namespace.
+
+---
+
+### Service and ServiceConfiguration Definitions
```
scotwells, line 193: I don't think it's important to have the actual service configuration file here. That's an implementation detail of registering the metering definitions.

scotwells (changes requested):
Directionally the architecture looks good. Let's just clean up some of the implementation details. We want to keep this document product / consumer focused.

scotwells:
Following on from the pipeline shape — the piece worth nailing down is how the log line gets the metadata embedded in the first place, since the project isn't something Envoy knows on its own. Project identity lives on the namespace (`resourcemanager.miloapis.com/project-name`, e.g. `p-abc`), so it has to be put onto the route before the access log can emit it.

The natural place to do that is the extension server proposed in #187. It already injects per-route `filter_metadata` for the WAF and Connector, already watches namespaces in a warm cache, and already maps the downstream `ns-<uid>` back to the upstream namespace — so adding the project name to route metadata is a small, additive change that stays out of the request path. Once it's on the route, the access log format references it with `%METADATA(ROUTE:...)%` and the line comes out with `project` (and whatever else we want) already embedded, ready for Vector to map to `subject: projects/{name}`.

So rather than enrich anywhere downstream, we piggyback on #187 to stamp the metadata at build time and let it fall through into the log line.

JoseSzycho:
@scotwells I just updated the enhancement, and I have a working implementation here:
- https://github.com/datum-cloud/network-services-operator/pull/188

Could you please review the Design Details I introduced in the previous commit?

I propose to different alternatives.

Will wait to hear from you in order to continue with the implementation.

docs/enhancements/http-traffic-metering.md
```diff
@@ -0,0 +1,496 @@
...
+components must be updated, all operating at the edge:
+
+1. **Network Services Operator (controller).** When the operator reconciles a
+   customer `HTTPRoute` into its downstream representation, it injects the
+   project name as a request header (`x-datum-project-name`) via a
+   `RequestHeaderModifier` filter on each route rule. The project name is read
+   from the upstream cluster identity (the Milo project name) that the
+   operator already holds while mapping upstream → downstream resources. Routes
+   that already define a `RequestHeaderModifier` are merged into rather than
+   duplicated, since Gateway API permits at most one such filter per rule.
```
scotwells: While this approach works, I believe the approach in https://github.com/datum-cloud/network-services-operator/pull/181#issuecomment-4721097933 would be better since it doesn't require exposing billing metadata through headers seen by end users. 

With the header approach, we would have to prevent the user from ever specifying this header in their request since we use it for billing attribution. Otherwise we could potentially bill the wrong project for an HTTP route.

scotwells (changes requested):

docs/enhancements/http-traffic-metering.md
JoseSzycho: thanks for the fast review. Will set this PR on draft and wait till the `extension server` to be on place for doing the final tweaks into the documentation.

JoseSzycho: @scotwells the extension servers works correctly. I followed the approach you recommended in https://github.com/datum-cloud/network-services-operator/pull/181#issuecomment-4721097933, and was able to correctly inject the project_name in the metadata.

Now the project_name is available in the per request access_logs. 

This PR https://github.com/datum-cloud/network-services-operator/pull/188 illustrates this. I crated the manifest for deploying the `extension server` in a dev environment, and was able to perform the e2e integration test for verifying the the updated `Design Details` described in this commit https://github.com/datum-cloud/network-services-operator/pull/181/commits/320d46a49ff7be76af22c6d8d3f6b80a5fa40ea2

Will wait for your review so I can proceed with the complete implementation of metering the HTTP Usage.

scotwells (approved):
Direction looks good. I would drop technical / implementation details like file paths, specific variables / strings, and the service configuration resources from the document. Those are all implementation details that are better defined in the actual code that's implemented to achieve this.

PR #187, opened 2026-06-16
Enhancement proposal: Envoy Gateway extension server

scotwells:
## Summary

Proposes replacing NSO's per-gateway **EnvoyPatchPolicy** objects with a single, secure **Envoy Gateway extension server** that applies Datum's Traffic Protection (WAF) and Connector customizations inline during config build.

This PR contains **only the enhancement proposal** (`docs/enhancements/envoy-gateway-extension-server/`) and its embedded diagrams — no code changes.

## Why this matters (customer impact)

- **New HTTPS gateways fail to get TLS certificates at scale.** Because all customer gateways share one Envoy configuration, every change rebuilds the whole fleet's config and re-applies every patch. At ~140 gateways this exceeds a minute and races Let's Encrypt's HTTP-01 challenge window — so customers see unexpected TLS errors on brand-new gateways.
- **WAF and Connector ride on fragile, after-the-fact patches** coupled to Envoy Gateway's flag-gated internal naming; when that coupling breaks, enforcement can fail silently.

## What we're proposing

A companion extension server, called **once per build** instead of N times, that injects the same WAF and Connector configuration through Envoy Gateway's supported extension hooks. Preliminary measurements show the build step dropping from ~72 ms to ~3 ms at N=10 gateways.

Customer outcomes:

- New HTTPS gateways get their certificate on the **first attempt**.
- WAF enforcement and Connector fallback behavior are **unchanged for customers** — only the internal delivery mechanism changes — and become **more robust** (no dependency on internal naming flags).

## Scope

- **No customer-facing API change.** Traffic Protection and the Connector are configured exactly as before.
- The proposal targets **Envoy Gateway v1.8.1** and details the security model (mTLS, least-privilege RBAC, network isolation), deployment topology, high availability, and an end-to-end verification plan.

## Reviewing

Start at `docs/enhancements/envoy-gateway-extension-server/README.md`. The proposal references the reference implementation and test harness that live under `test/perf/`; those land in follow-up PRs, so they are cited by path rather than linked (this PR links only to files already on `main`).


slindseysr (dismissed):

savme (dismissed):

ecv (dismissed):
If _extension servers_ work as advertised then they're such an ideal solution that it makes me suspicious they were invented by someone at Envoy who ran into this exact problem.

0xmc (dismissed):
@scotwells you addressed all my concerns in Non-Goals, Risks and Mitigations, and Drawbacks.

scotwells:
Going to leave this unmerged until I can validate the architecture.

scotwells:
Had to make one change to the design to trigger the envoy gateway to trigger re-translation of the Envoy configuration when a TrafficProtectionPolicy or a Connector resource is updated. Changes are captured in https://github.com/datum-cloud/network-services-operator/pull/187/commits/f0fd9dc8841702d1ccc302ac8c0446ebddd5f82e.

mattdjenkinson (approved):

scotwells:
Cross-linking the implementation: #189 lays the groundwork for this proposal by getting Envoy Gateway and Gateway API onto the versions the extension server needs. The work described here builds on top of it.

PR #188, opened 2026-06-17
Feat/envoy logs

JoseSzycho:
>[!WARNING]
> This PR should not be merged. It has been created to show how we can capture the HTTPRoute usage signals.
> Once the corresponding enhamcenete has been approved and merged into main, continue with the work on this PR.

Related to:
- https://github.com/datum-cloud/network-services-operator/pull/181

### Summary

The billing pipeline previously could only attribute usage to a project by its control-plane namespace UID (ns-<uid>). This PR surfaces the human-readable project name as a first-class dimension on every billing CloudEvent, enabling clearer cost attribution without post-processing joins.

- Extension server (xDS mutation): ApplyTPPRouteConfig now calls injectProjectNameMetadata on every NSO-owned route, stamping project_name into filter_metadata["datum-gateway"]["project_name"] on the Envoy route proto at xDS build time. This happens unconditionally — WAF config is an optional overlay on top.
- Envoy access log format: the EnvoyProxy JSON format is extended with project_name: "%METADATA(ROUTE:datum-gateway:project_name)%", read directly from the xDS route metadata on every request.
- Vector billing collector: adds a kubernetes_logs source (scoped to envoy-gateway-system) and a project_name extraction in the VRL transform, emitting it as a dimension on all four billing CloudEvents (requests, ingress-bytes, egress-bytes, connection-seconds).
- E2E test: configures the EnvoyProxy with a File sink only (matching the production datum-downstream-gateway), asserts the project_name dimension in the received CloudEvents.

### Notes

- %METADATA(ROUTE:datum-gateway:project_name)% is used because the project name lives in xDS route metadata injected by the extension server — it is not a per-request value and does not need to travel as a header.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,8 @@ test-e2e: chainsaw
 	}
 	$(KIND) get kubeconfig --name nso-standard > $(TMPDIR)/.kind-nso-standard.yaml
 	$(KIND) get kubeconfig --name nso-infra > $(TMPDIR)/.kind-nso-infra.yaml
-	$(CHAINSAW) test ./test/e2e \
+	$(CHAINSAW) test $(or $(TEST_DIR),./test/e2e) \
+	    --parallel 1 \
 		--cluster nso-standard=$(TMPDIR)/.kind-nso-standard.yaml \
 		--cluster nso-infra=$(TMPDIR)/.kind-nso-infra.yaml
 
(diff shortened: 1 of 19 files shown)


JoseSzycho:
replaced by https://github.com/datum-cloud/network-services-operator/pull/239

PR #205, opened 2026-06-19
feat(extension-server): branded data-plane error pages

scotwells:
## Why

When a customer's site is reached through Datum's edge and the origin can't be served — an offline Connector tunnel, a down backend, a timeout — Envoy answers, not the customer's app. On the downstream/Connector data plane that's a raw `503 no healthy upstream`. This serves a **Datum-branded HTML error page** there instead — the page a customer's own users see during the customer's outage.

Implements the **Branded Data-Plane Error Pages** enhancement — proposal in #204 ([doc](https://github.com/datum-cloud/network-services-operator/blob/docs/branded-error-pages-enhancement/docs/enhancements/branded-error-pages/README.md)).

## What changed

- New `InjectLocalReplyConfig` listener mutation in the extension server: any edge-generated response ≥ 500 renders the branded body (status code preserved). Covers HTTP/2 and HTTP/3 listeners; skips internal/readiness listeners; idempotent.
- Content from an **optional mounted ConfigMap** (`envoy-error-pages`) with a **compiled-in fallback** — works out of the box, editable later via GitOps.
- Config: `gateway.errorPage.{enabled,bodyPath,minStatusCode}`.

## Notes for review

- **Fail-safe**: the downstream hook is `failOpen:false`; missing/empty content is a no-op, and the injector errors only on a malformed HCM.
- Ships working via the embedded default. The `envoy-error-pages` ConfigMap content + a content-change reload are a **follow-up in the infra repo**.
- `golangci-lint` skipped locally (go1.25 vs repo's go1.26.4 toolchain — not a code issue).

Draft — not for merge/release yet; pending live verification on staging (503 → branded HTML).


scotwells:
@mattdjenkinson we'd keep the real template we use in the infra repo so we don't need to do software releases to update the template

mattdjenkinson (approved):

PR #211, opened 2026-06-23
Fix tunnels staying offline at the edge after a connector connects

### LINKED ISSUE #209: Connector upstream-status annotation not re-mirrored after Ready condition flips True
## Summary

When a tunnel connector's `Ready` condition transitions `False → True` (heartbeat renews the lease), the downstream connector's `networking.datumapis.com/upstream-status` annotation is not updated. The extension server reads this stale annotation and classifies the connector as offline, causing persistent HTTP 503 responses.

## Sequence

1. Connector created — replicator mirrors status into the downstream annotation. At this moment `Ready: False` (lease not yet renewed).
2. Heartbeat connects — upstream connector becomes `Ready: True`, `connectionDetails` populated.
3. Replicator does **not** re-reconcile — `skipUpstreamStatusSync: true` suppresses downstream→upstream sync, and there is no watch on upstream *status* changes to re-trigger the mirror path.
4. Downstream connector's annotation still contains `Ready: False` with stale `connectionDetails`.
5. Extension server reads the annotation, sees `Ready: False`, marks all routes for that connector as offline → HTTP 503.

## Evidence

Downstream connector annotation captured while upstream connector was `Ready: True`:

```json
{
  "conditions": [
    { "type": "Ready", "status": "False", "reason": "ConnectorNotReady",
      "message": "Connector lease has expired. Agent may be offline." }
  ],
  "connectionDetails": { "publicKey": { "id": "378843c806c8c93c5770abaa19bc47e04e9f56977c6e7cc28044a09ef5a1cd23", ... } }
}
```

Upstream connector at the same moment:

```json
{ "type": "Ready", "status": "True", "reason": "ConnectorReady",
  "message": "The connector is ready to tunnel traffic." }
```

Extension server log confirms all vhosts for this connector remain offline:
```
connector_offline_routes:26  clusters_replaced:1
```

## Root Cause

`replicationResourceConfig` for the Connector type sets:

```go
mirrorStatusToAnnotation: true,
skipUpstreamStatusSync:   true,
```

The replicator reconciles on spec changes but has no watch/enqueue path for upstream *status* changes. When `Ready` flips after the initial replication, nothing re-queues the replicator to update the annotation.

`touchDownstreamGatewayAnnotations` in the connector controller fires to trigger EG re-translation, but the stale annotation means the extension server still classifies the connector as offline even after re-translation.

## Fix

The replicator should watch upstream connector status changes and re-mirror the annotation when status changes. Alternatively, the connector controller should enqueue a replicator reconcile after writing `Ready: True`.

scotwells:
## What this fixes

A tunnel connector could come online but the edge would keep serving **HTTP 503** for it — the data plane never picked up that the tunnel was ready. The interim workaround was a CLI re-patch to force it.

Fixes #209.

## Why it happened

The connector's live status reaches the edge correctly, but Envoy Gateway wasn't re-translating against it. EG only re-runs its translation — and the extension hook that programs tunnels — when a resource it watches changes through an annotation-aware path, and a `Connector` update doesn't qualify. So fresh liveness sat in the extension server's cache while EG kept serving the stale (offline) program.

## What changed

- A small controller co-located with the extension server watches the `Connector` and, when its liveness changes, nudges the owning `Gateway` so EG re-translates — after the fresh liveness is already in the local cache.
- Removed the earlier trigger that ran from the project control plane and raced the connector status's propagation to the edge.
- RBAC: the extension server may now `patch` Gateways.

## Impact

- Connectors come online at the edge on their own, within a few seconds of connecting — no manual step.
- The connect-lib `refresh_connection_details()` workaround is no longer required (safe to leave or remove).

## Testing

- Unit tests for the new controller (liveness → Gateway touch, offline value, missing Gateway, change-only triggering).
- Existing connector and replicator suites updated and passing.

## Follow-ups (not in this PR)

- Leader election for the new controller (it runs on every replica today; patches are idempotent).
- An e2e test for the edge re-translation path.


ecv (approved):
ok

PR #217, opened 2026-06-24
Don't propagate gateway listeners with invalid TLS certificates

### LINKED ISSUE #212: Expired/abandoned customer certs can freeze an entire edge — health-gate cert propagation + isolate per-cert failures
## Summary

A single customer's expired/abandoned TLS certificate can freeze **all** config propagation to an entire edge POP — for every tenant on that edge, not just the cert's owner. On 2026-06-23 we found **15 of 16 production edges** frozen this way, dating to ~2026-06-20; only `us-east-2` was healthy.

This is **not a one-off** — it's a normal customer lifecycle that will recur naturally:

```
customer moves a domain away from Datum
  → DNS stops pointing at us → ACME HTTP-01 renewal fails
  → cert-manager Certificate goes Ready=False, the cert expires
  → the broken Secret lingers and keeps getting re-propagated to edges
  → Envoy Gateway eventually serves a stale/"phantom" cert for it over SDS
  → that one bad cert/key NACKs the whole tcp-443/udp-443 listener (atomic update)
  → the entire POP's config is frozen on its last-good listener
```

Same failure family as #194 (a missing OIDC clientSecret takes down the shared downstream gateway) and #183: one tenant's bad input breaks the shared listener for everyone.

## Customer impact

While an edge is frozen, its downstream gateway listener is stuck on its last-good config:

- Cert renewals, new custom domains, route changes, and the branded error page **stop reaching that edge** — it silently serves stale config.
- The failure is **invisible**: Gateway status stays `Programmed=True` and no existing alert fired (`EnvoyListenerCreateFailure` stays at 0 in this scenario — the failure is a rejected *update*, `lds.update_rejected`, not a failed *create*).
- A data-plane proxy restart on a frozen edge **cannot reprogram the listener** → that POP goes offline.

## Root cause

1. **EG delivers a phantom cert over SDS.** For the affected secrets, EG serves a cert (`5A00F2…`) that exists in **no current secret** — a stale cached copy paired with a key that no longer matches (`Failed to load private key … KEY_VALUES_MISMATCH`). The *current* source secret is a consistent (if expired) cert/key pair that Envoy would load fine, so the freeze is the **EG SDS staleness**, not the expiry itself. A controller restart, a secret update, and even deleting the source secret do **not** evict the phantom from EG's cache.
2. **One bad cert NACKs the whole listener.** Because a listener update is atomic, a single unloadable filter chain rejects the entire `tcp-443`/`udp-443` listener for all tenants.

## Why the obvious workarounds don't hold

The cert reaches the edge via: customer CP → **NSO gateway-resource-replicator** (selects Secrets labeled `networking.datumapis.com/gateway-sync`, `internal/config/config.go:1110`) → federated CP → **Karmada `ClusterPropagationPolicy nso-resources`** (`conflictResolution: Overwrite`, infra `apps/network-services-operator/downstream/federated/clusterpropagationpolicy.yaml`) → edges. Deleting the Secret at the edge unfreezes it transiently, but the `Overwrite` policy **re-propagates it within ~80s** (cert-manager keeps retrying the failed renewal upstream, re-touching the Secret). So edge-side cleanup is a losing race; the fix must be upstream and structural.

## Proposed fix (defense in depth)

**1. Primary — health-gate the cert at the source NSO controls.** Today the replicator ships any `gateway-sync`-labelled Secret regardless of cert health (`gateway_resource_replicator_controller.go`, `internal/config/config.go:1110`). Replicate a per-hostname cert Secret **only while its cert-manager Certificate is `Ready=True` and unexpired**, and **remove** the federated copy when it goes not-Ready/expired. Karmada follows the source, so the edge copy is pruned durably (no `Overwrite` fight). This automates the manual fix on the real trigger and directly handles the offboarding lifecycle.

   - Behavior change to confirm: an offboarded domain's listener simply drops once its cert dies, instead of serving an expired cert. (It was already broken for that customer — but worth an explicit product decision.)
   - Implementation note: confirm the replicator deletes the downstream copy when a resource stops matching (not just create/update).

**2. Backstop — one bad cert must not NACK the whole listener.** Even with (1), isolate per-cert failures so *any* bad cert (EG phantom, torn rotation, malformed upload) drops only its own SNI chain. **Feasible in the ext-server `PostTranslateModify`:** it already receives the secrets (`req.GetSecrets()`, currently passed through unchanged at `internal/extensionserver/server/server.go:288`) and already walks every filter chain (Coraza + LocalReply injection). Extend it to validate each secret's cert/key (parse, match, expiry) and prune the bad secret + its filter chains before returning, so the listener always applies.

**3. Detection.** Alert when an edge listener is rejecting updates (frozen). Rule drafted in infra (`EnvoyListenerUpdateRejected`: `lds.update_rejected` rising with zero `lds.update_success`); promtool-clean.

**4. (Optional, upstream) EG SDS staleness.** Investigate/fix why EG serves a phantom cert that survives a controller restart and source-secret deletion. (1)+(2) make this non-urgent, but it's the deepest root cause.

## Workaround (validated, non-durable)

Delete the offending expired cert Secret(s) on the affected edge **and** restart the EG controller; the listener re-applies and config (incl. the branded page) reaches the data plane. Validated on us-west-1 2026-06-23 — but Karmada re-propagates the Secret, so it does not hold without the structural fix above.

scotwells:
Fixes #212

## Problem
Every HTTPS hostname on a gateway shares a single Envoy `:443` listener. When one customer's certificate becomes unusable — expired, or no longer issuable because the domain moved away from Datum — Envoy rejects the **entire** listener update, so HTTPS goes down for **every tenant on that edge**, not just the affected hostname. The failure was also silent: nothing told the customer their certificate was the problem, and nothing alerted operators.

### Isolate bad certificates and tell the customer
- A listener whose certificate is expired, not-yet-valid, missing, mismatched, or not yet issued is withheld from the downstream gateway. A bad certificate now only affects its own hostname; every other hostname keeps serving.
- The listener reports standard `Programmed` / `ResolvedRefs` conditions with concise, non-technical messages (e.g. *"The TLS certificate for example.com has expired, so HTTPS for this hostname is paused…"*). It stays `Accepted` — the config is valid, the certificate just isn't usable yet.
- **Data-plane backstop:** if a bad certificate ever reaches the edge anyway, the extension server drops just that one part of the listener instead of letting it reject the whole listener.
- **Self-healing:** once the certificate can be issued, the listener is propagated and serves automatically.

### Make it observable
Previously this gating was invisible to operators. This PR adds the full chain from signal to action:
- **Metrics:** per-listener gauges for withheld listeners, certificate expiry time, and managed listeners, a gating counter, and extension-server gauges for the backstop's live state. Labelled by gateway/listener/hostname so an operator can query the exact dark hostname during an incident; series are cleared on recovery, listener removal, and gateway deletion so a stale value never reports a phantom dark hostname.
- **Alerts** (`config/telemetry/alerts`): a customer listener withheld for a bad certificate, a certificate expiring soon (act before it gates), the backstop actively dropping certificates, and — critical — a listener that can't be protected at all and will have its edge update rejected. They complement the infra `EnvoyListenerUpdateRejected` alert.
- **Runbooks** (`docs/runbooks/gateway-tls-certificates.md`): each alert links via `runbook_url` to a section with meaning, impact, diagnosis, and remediation.
- **Scrape wiring:** ServiceMonitors so the operator and extension-server metrics actually reach Prometheus (otherwise the alerts could never fire).

## How it's verified
- **Unit tests** for the certificate-health gate and the data-plane prune.
- **End-to-end (chainsaw) test** `test/e2e/gateway-invalid-certificate` on a real two-cluster + Envoy Gateway setup: the bad listener is withheld while the good sibling is unaffected; the user-facing status message is shown; a **real HTTPS request** to the good hostname returns `200` while the bad hostname is not served; and after the certificate issues, the listener reappears and serves `200`.
- **Negative control:** with the fix reverted, the e2e test fails at the omission assertion — confirming it actually guards the behavior.
- **promtool unit tests** for all four alert rules.
- CI green (unit, e2e, lint, kustomize).

## Notes
- `goconst` is disabled in a separate commit: for log fields and map keys like `"reason"`/`"message"` the inline literal is clearer than a shared constant, and forcing one couples unrelated call sites.
- Runbooks were added for the certificate alerts this PR introduces; the pre-existing SLO / EnvoyPatchPolicy alerts in the same file are left as a separate follow-up.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,7 +8,6 @@ linters:
     - dupl
     - errcheck
     - ginkgolinter
-    - goconst
     - gocyclo
     - govet
     - ineffassign
@@ -37,11 +36,6 @@ linters:
           - dupl
           - lll
         path: internal/*
-      # Repeated string literals in tests are usually fixture/table data;
-      # extracting them to constants hurts readability more than it helps.
-      - linters:
-          - goconst
-        path: _test\.go
       # The validation packages are built almost entirely from field.ErrorList
       # accumulators that hold a handful of errors; preallocating them adds noise
       # without meaningful benefit.
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -17,6 +17,13 @@ namePrefix: network-services-operator-
 components:
 - ../resource-metrics
 - ../webhook
+# Prometheus ServiceMonitor for the controller-manager metrics endpoint.
+# The controller serves metrics over HTTPS on :8443 with delegated authn/authz
+# (controller-runtime WithAuthenticationAndAuthorization). The ServiceMonitor
+# uses insecureSkipVerify because the controller auto-generates a self-signed
+# TLS cert — there is no cert-manager-issued cert for the metrics endpoint and
+# no CA bundle to reference. Prometheus still authenticates via the bearer token.
+- ../prometheus
 
 resources:
 - ../crd
@@ -26,10 +33,7 @@ resources:
 # crd/kustomization.yaml
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
 #- ../certmanager
-# [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
-#- ../prometheus
-# [METRICS] Expose the controller manager metrics service.
-- metrics_service.yaml
+# metrics_service.yaml is now included by the ../prometheus component above.
 # [NETWORK POLICY] Protect the /metrics endpoint and Webhook Server with NetworkPolicy.
 # Only Pod(s) running a namespace labeled with 'metrics: enabled' will be able to gather the metrics.
 # Only CR(s) which requires webhooks and are applied on namespaces labeled with 'webhooks: enabled' will
--- a/config/extension-server/kustomization.yaml
+++ b/config/extension-server/kustomization.yaml
@@ -9,6 +9,7 @@ resources:
   - rbac
   - certmanager
   - network-policy
+  - metrics-monitor.yaml
 
 images:
   - name: ghcr.io/datum-cloud/network-services-operator
--- a/config/extension-server/metrics-monitor.yaml
+++ b/config/extension-server/metrics-monitor.yaml
@@ -1,0 +1,22 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/name: network-services-operator
+    app.kubernetes.io/component: envoy-gateway-extension-server
+    app.kubernetes.io/managed-by: kustomize
+  name: envoy-gateway-extension-server-metrics
+  namespace: system
+spec:
+  endpoints:
+    # The extension server serves /metrics on plain HTTP (no TLS) on the
+    # health-addr port (:8080). Only the gRPC port uses mTLS; the health
+    # address intentionally stays plain HTTP so Kubernetes probes don't
+    # need certificates.
+    - path: /metrics
+      port: metrics
+      scheme: http
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: network-services-operator
+      app.kubernetes.io/component: envoy-gateway-extension-server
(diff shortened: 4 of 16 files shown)


mattdjenkinson (approved):

scotwells:
Confirmed connector's came online correctly after this was deployed to staging.

PR #220, opened 2026-06-24
fix: clear error for hostname-less tenant listeners

### LINKED ISSUE #219: Gateway: hostname-less tenant listeners rejected with misleading 'must be unique' (default-listener injection collision)
## Summary

A tenant `Gateway` (class `datum-external-global-proxy`) whose listeners use the default ports/protocols (80/HTTP, 443/HTTPS) **without a `hostname`** is rejected at admission with a misleading upstream message:

```
Gateway.gateway.networking.k8s.io "http-gateway" is invalid: spec.listeners:
  Invalid value: Combination of port, protocol and hostname must be unique for each listener
```

The spec is valid per upstream Gateway API (listener `hostname` is optional). The rejection comes from NSO's default-listener injection colliding with the tenant's listeners.

## Reproduction (live, `kubectl apply --dry-run=server`)

| # | Listeners | Hostname(s) | Result |
|---|-----------|-------------|--------|
| A | 2 — `80/HTTP`, `443/HTTPS` | none | ❌ "must be unique" |
| B | **1** — `80/HTTP` | none | ❌ "must be unique" |
| C | 2 — `80/HTTP`, `443/HTTPS` | distinct | ✅ |
| D | **1** — `80/HTTP` | set | ✅ |
| E | 2 — `80/HTTP`, `443/HTTPS` | **same** hostname | ✅ |

Probe **B** is decisive: a single tenant listener cannot collide with itself, yet it is rejected — so the failure is not tenant-side tuple duplication.

## Root cause

`SetDefaultListeners` (`internal/util/gateway/listeners.go`) unconditionally appends two listeners, **both without a `hostname` at creation**:

- `default-http` → `80 / HTTP`
- `default-https` → `443 / HTTPS`

`IsDefaultListener` keys off the listener **name** (`default-http` / `default-https`) only. When a tenant defines their own `80/HTTP` and/or `443/HTTPS` listeners under different names and omits `hostname`, defaulting produces duplicate `(port, protocol, hostname=∅)` tuples — e.g. tenant `http` (80,HTTP,∅) + injected `default-http` (80,HTTP,∅). The upstream Gateway API CRD CEL then rejects the whole object with `Combination of port, protocol and hostname must be unique for each listener`.

Probes C/D/E pass because giving the tenant listeners a `hostname` makes their tuple distinct from the hostname-less injected defaults.

Separately, `validateListeners` (`internal/validation/gateway_validation.go:72`) already requires a `hostname` on non-default listeners (`must be set to "…" or a custom hostname`). So hostname-less tenant listeners are unsupported by design — but the tenant never sees that clear message; the injected-default collision trips the CEL first and surfaces the confusing "must be unique" error instead.

## Impact

- Misleading error: tenants see a uniqueness complaint about listeners they didn't duplicate, with no indication that defaults were injected.
- Blocks `datum-cloud/infra` integration test `tests/networking/http-gateway` (datum-cloud/infra#2804), whose Gateway defines plain `http`/`https` listeners with no hostname.

## Suggested fixes (pick per design intent)

1. Make defaulting idempotent against tenant listeners — skip injecting `default-http`/`default-https` when a tenant listener already covers that port/protocol (dedupe by `(port, protocol)`, not just name).
2. Set the expected DNS hostname on the injected default listeners at creation so their tuples are always distinct.
3. If hostname-less listeners are simply unsupported, fail fast in `validateListeners` with the clear `hostname is required` message **before** defaulting can produce a colliding tuple — so the surfaced error is actionable.

## Refs

- datum-cloud/infra#2804 (test failure + full investigation matrix)

ecv:
## Summary

Closes #219.

Tenant Gateways (class `datum-external-global-proxy`) whose listeners use the managed ports (80/443) **without a hostname** were rejected at admission with the opaque upstream CEL message:

```
spec.listeners: Invalid value: Combination of port, protocol and hostname must be unique for each listener
```

— even for a *single* listener, which can't collide with anything.

## Root cause

The mutating defaulter (`GatewayCustomDefaulter.Default`) injects `default-http` (80/HTTP) and `default-https` (443/HTTPS) with **no hostname at creation** (the canonical `<uid>.<targetDomain>` hostname is stamped later by the controller, since the UID isn't available in the create webhook). A tenant listener on the same port/protocol without a hostname produces a duplicate `(port, protocol, hostname=∅)` tuple, which the upstream Gateway API CRD CEL rejects.

Admission order is **mutating webhook → CRD schema/CEL validation → validating webhook**, so the CEL fires before `validateListeners` runs. That makes the existing `validateListeners` hostname-required branch (`gateway_validation.go`) effectively unreachable for tenant listeners at creation — the opaque CEL error always wins.

## Fix

Reject hostname-less tenant (non-default) listeners in the **defaulting webhook**, before injecting the defaults — the defaulter returns `error`, which denies admission *before* the CEL stage, so the operator surfaces an actionable message:

```
spec.listeners[0].hostname: Required value: a hostname must be set; this controller
injects default HTTP/HTTPS listeners, so listeners you define must specify a
distinguishing hostname
```

`default-`named listeners are exempt (they're overwritten by defaulting and stamped by the controller), and bare gateways (no listeners) are unaffected.

This is the design-intent-aligned option from the issue discussion (#3): the design already mandates a distinguishing hostname on tenant listeners (ports are locked to 80/443, so hostname is the only uniqueness axis), and the injected defaults are meant to coexist with hostname-bearing tenant listeners. It does **not** dedupe/suppress the default listeners (#1, would drop the canonical hostname) and does not try to stamp a hostname at create (#2, UID unavailable then).

## Tests

- `TestValidateTenantListenerHostnames`: bare gateway, hostname-less tenant listener (rejected), hostname-bearing tenant listener (ok), `default-`named listeners (exempt).
- `go build ./...`, `go vet`, and the validation package tests pass.

## Validation note

Reproduced the original failure against a live project control plane with `kubectl apply --dry-run=server`: a single 80/HTTP listener with no hostname was rejected with the CEL "must be unique" message; adding a hostname made it apply cleanly — confirming the diagnosis.

Refs datum-cloud/infra#2804.

### DIFF
--- a/test/e2e/gateway-listener-hostname-collision/chainsaw-test.yaml
+++ b/test/e2e/gateway-listener-hostname-collision/chainsaw-test.yaml
@@ -1,39 +1,44 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
 #
 # Issue #219 — a tenant Gateway whose listeners use the default ports/protocols
-# (80/HTTP, 443/HTTPS) WITHOUT a hostname is rejected at admission with the
-# misleading upstream message:
+# (80/HTTP, 443/HTTPS) WITHOUT a hostname used to be rejected at admission with
+# the misleading upstream CEL message:
 #
 #   Combination of port, protocol and hostname must be unique for each listener
 #
-# The spec is valid per upstream Gateway API (listener hostname is optional).
-# The rejection comes from NSO's default-listener injection: the mutating
-# webhook unconditionally appends hostname-less default-http (80/HTTP) and
-# default-https (443/HTTPS) listeners. When a tenant defines their own 80/HTTP
-# or 443/HTTPS listener and omits the hostname, defaulting produces duplicate
-# (port, protocol, hostname=empty) tuples, and the upstream CRD CEL rejects the
-# whole object before NSO's own validating webhook can surface a clear reason.
+# Root cause: NSO's default-listener injection. The mutating webhook
+# unconditionally appends hostname-less default-http (80/HTTP) and default-https
+# (443/HTTPS) listeners. When a tenant defines their own 80/HTTP or 443/HTTPS
+# listener and omits the hostname, defaulting produced duplicate
+# (port, protocol, hostname=empty) tuples, and the upstream CRD CEL rejected the
+# whole object before NSO's own validating webhook could surface a clear reason.
 #
-# This is a pure admission-time reproduction: the rejection happens on the
+# The fix (#220) rejects hostname-less tenant (non-default) listeners in NSO's
+# defaulting webhook, which runs BEFORE the CRD CEL stage. The object is still
+# rejected — but now with an actionable message:
+#
+#   spec.listeners[N].hostname: Required value: a hostname must be set; this
+#   controller injects default HTTP/HTTPS listeners, so listeners you define
+#   must specify a distinguishing hostname
+#
+# This test now asserts the FIXED behavior: probes A/B are still rejected, but
+# with the clear hostname-required message instead of the opaque "must be
+# unique" CEL error. Probe D (listener WITH a hostname) is still admitted.
+#
+# This is a pure admission-time check: the rejection happens on the
 # control-plane (nso-standard) cluster, so no downstream/edge wiring is needed.
 # We drive it exactly the way the issue did — kubectl apply --dry-run=server —
 # which runs the mutating webhook + CRD CEL without persisting anything.
 #
 # The probes mirror the issue's investigation matrix:
-#   B (decisive) — 1 listener, 80/HTTP, no hostname            -> rejected
-#   A            — 2 listeners, 80/HTTP + 443/HTTPS, no hostname -> rejected
+#   B (decisive) — 1 listener, 80/HTTP, no hostname            -> rejected (clear msg)
+#   A            — 2 listeners, 80/HTTP + 443/HTTPS, no hostname -> rejected (clear msg)
 #   D (control)  — 1 listener, 80/HTTP, WITH a hostname         -> accepted
 #
 # Probe B is decisive: a single tenant listener cannot collide with itself, yet
-# it is rejected — proving the failure is the injected-default collision, not
-# tenant-side duplication. Probe D is the positive control: adding a hostname
-# makes the tuple distinct from the injected defaults and the object is admitted.
-#
-# Run against unfixed code (main / this branch), all three probes pass: the bug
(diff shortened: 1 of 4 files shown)


ecv:
Setting to draft until the bug this fixes is reproduced with end to end tests in this repo.

ecv:
E2E reproduction of the bug this PR fixes: #221 (separate branch, test-only, no production code). It reproduces the misleading `must be unique` rejection on unfixed `main`. When this fix merges, probes A/B in that test will need their expected message updated to the clear hostname-required message.

ecv:
Closed accidentally by GitHub automation tripping up on a 'closes' verb in a related issue without actually understanding English.

ecv:
Fixing tests.

ecv:
**Heads up on the E2E failure — it was expected, now fixed.**

The red check was `chainsaw/gateway-listener-hostname-collision`, Probe B. That test (merged separately in #221) was written to *reproduce* #219: it asserted that a hostname-less tenant listener gets rejected with the opaque upstream CEL message:

```
Combination of port, protocol and hostname must be unique for each listener
```

This branch *fixes* #219 by rejecting those listeners in NSO's defaulting webhook — which runs **before** the CRD CEL stage — so the object is still rejected, but now with an actionable message:

```
spec.listeners[0].hostname: Required value: a hostname must be set; this
controller injects default HTTP/HTTPS listeners, so listeners you define
must specify a distinguishing hostname
```

Object still denied (correct), but the error string changed, so the probe's `grep` for "must be unique" failed. The repro test even called this out in its own header comment ("...so this test's expected message will change with the fix").

Fixed in the latest push:
- Probes A/B now `grep` for `a hostname must be set` and are reframed from "reproduce the bug" to "verify the fix".
- Probe D (listener *with* a hostname → admitted) unchanged.
- Also merged `main` into the branch — the repro test lived only on `main`, so CI was running it via the merge ref while the branch itself didn't carry the file.

CI is re-running; Probes A/B should go green.

ecv:
ok @scotwells for real this time. how's my driving

happy to rip those comments out btw, this repo lacks my CLAUDE.md strictness (adding that separately)

ecv:
cc #229 for CLAUDE.md

test/e2e/gateway-listener-hostname-collision/chainsaw-test.yaml
scotwells, line 1: I don't see anything actually changed in this test? It all looks like comment changes?

scotwells: Oh, I guess just the error being returned is different.

scotwells (approved):

PR #224, opened 2026-06-25
Let the edge report the protections it was told to run

scotwells:
## What this enables

The edge applies firewall and connector protections on behalf of the platform — but until now there was no way to ask the edge what protections it believes it should be running.

This adds a small, read-only endpoint that reports exactly that: the set of rules the platform handed to the edge. On its own it changes nothing a customer can see. Its value comes next — it's the reference point that lets us confirm the edge is actually running what it was told to, instead of assuming that a healthy-looking response means the configuration took effect.

## Why it matters

A request can succeed while the protection behind it never applied. Without a way to ask the edge what it intends to enforce, that gap stays invisible. This is the first half of closing it.

## Safety

Read-only and off to the side of the traffic path — it observes configuration, it never touches live requests.

---

**Part of #247** — catching silent edge misconfiguration before it reaches customers.

One of a five-PR stack, best reviewed and merged in order:

- **#224 — report the protections the edge was told to run ← you are here**
- #225 — confirm the edge is running the configuration it was given
- #226 — a test environment that mirrors how the edge really runs
- #227 — carry edge policy from the control plane out to the edges
- #228 — prove the edge's core guarantees with real traffic

### DIFF
--- a/config/extension-server/deployment.yaml
+++ b/config/extension-server/deployment.yaml
@@ -58,6 +58,7 @@ spec:
             - --tls-key=$(TLS_KEY)
             - --tls-client-ca=$(TLS_CLIENT_CA)
             - --server-config=$(SERVER_CONFIG)
+            - --enable-programmed-set=$(ENABLE_PROGRAMMED_SET)
           env:
             - name: GRPC_ADDR
               value: ":5005"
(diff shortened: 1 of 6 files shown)


ecv:
i'd start to merge these if yer game

ecv (approved):

PR #226, opened 2026-06-25
Stand up a test environment that mirrors how the edge really runs

scotwells:
## What this stands up

Our edge tests have run against a simplified setup that didn't match how the edge actually runs in production. Whole classes of problems — the ones that only appear with the real gateway, the real firewall data plane, and the real path the platform uses to hand configuration to the edge — could never show up in a test.

This brings up a test environment that mirrors the real edge: two clusters wired the way production is, the same gateway version we run in production, the firewall data plane customer traffic actually passes through, and the same delivery path configuration really travels. Tests can now exercise the route real traffic takes, instead of a stand-in for it.

## Why it matters

A test is only as trustworthy as how closely it resembles the real thing. Every gap between the test setup and production is a place a bug can hide unseen. Closing that gap is what makes the end-to-end tests built on top of this worth trusting.

## Scope

This is test scaffolding — it changes nothing about how the edge runs for customers. It only gives us a faithful place to prove the edge behaves before changes reach them.

---

**Part of #247** — catching silent edge misconfiguration before it reaches customers.

One of a five-PR stack, best reviewed and merged in order:

- #224 — report the protections the edge was told to run
- #225 — confirm the edge is running the configuration it was given
- **#226 — a test environment that mirrors how the edge really runs ← you are here**
- #227 — carry edge policy from the control plane out to the edges
- #228 — prove the edge's core guarantees with real traffic

### DIFF
--- a/Taskfile.test-infra.yml
+++ b/Taskfile.test-infra.yml
@@ -1,0 +1,419 @@
+version: '3'
+
+# Production-fidelity test environment. Two kind clusters: nso-upstream (the
+# control plane) and nso-downstream (the edge). Consumed by both the e2e and perf
+# suites.
+#
+# Bring up:   task test-infra:up
+# Tear down:  task test-infra:down
+# Smoke:      task test-infra:smoke
+
+# kustomize fetches some CRDs over https. A developer's global git config can
+# rewrite https to ssh, which fails in a non-interactive fetch. Neutralize the
+# global git config for kustomize's child git so the https fetches succeed; the
+# repo-local config is untouched.
+env:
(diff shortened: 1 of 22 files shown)


ecv:
Superseded by #276 — this env rebased onto `main` plus the e2e suite (#275) lifted onto it, as a single PR for this slot in the #247 stack. The suite runs 8/8 green against the live two-cluster env there. Leaving this open for you to close/redirect as you see fit, @scotwells.

ecv:
@scotwells permission to close in favor of #276? – also, lets start mergin the stack

PR #230, opened 2026-06-27
fix: gate ActivityPolicy create/update audit rules on 2xx outcome

ecv:
## What

Gate every **create** and **update** audit rule across all NSO ActivityPolicies on the request outcome — append `audit.responseStatus.code >= 200 && audit.responseStatus.code < 300` to each rule's `match` — so an activity rule fires only for a request that actually succeeded. Ships a `test/activitypolicy` suite that proves the invariant.

Consolidates the two changes that converged on this mechanism (this branch + #241) plus the tests into one PR.

## Why — one defect, two symptoms

Create/update rules matched on `has(audit.requestObject.spec)` (or the verb alone), so a **rejected** request (403 / 409 / 422 / admission-deny) still matched. Two failures fall out of that:

1. **DLQ leak.** On a rejected write the apiserver returns a `Status` object as `responseObject` — no `spec`, no `metadata.name`. A summary dereferencing `audit.responseObject.<leaf>` throws, the event is lost to the DLQ, and retries fail identically. Prod `DLQSlowLeak` on `gateway.networking.k8s.io-gateway` (`error_type=cel_summary`, ~120 events/6h, climbing):

   ```
   rule 0 summary: summary template evaluation failed:
   eval "link(audit.responseObject.metadata.name, audit.objectRef)": no such key: name
   ```

2. **False activities.** Even rules that dereference only `requestObject` / `objectRef` (so they don't DLQ) still emit a false "created X" / "updated X" in the timeline for an attempt that never succeeded.

Gating on 2xx fixes both at once — the rule simply does not fire on a failed request.

## Scope

All 8 policies — `backendtlspolicy`, `connector`, `connectoradvertisement`, `domain`, `gateway`, `httpproxy`, `httproute`, `trafficprotectionpolicy` — for `create`, `create-fallback`, `update`, `update-fallback`.

`delete` rules are unchanged (they already guard `has(audit.responseObject.spec)`; delete responses are not outcome-gated here).

The `httpproxy` / `httproute` **update** rules were the last gap — the create-gate change had missed them — and were surfaced by the new test, then closed.

## Tests (`test/activitypolicy`)

- **structural** — every create/update rule's `match` gates on `responseStatus.code` in `[200,300)`
- **semantic (cel-go)** — each create/update rule matches a successful write and does **not** match a failed one
- **compile** — every rule `match` parses

## Prod remediation

Merge + NSO release → Flux re-applies the corrected ActivityPolicy CRs (overwriting the live objects); the processor retries the DLQ backlog. No `kubectl`.

## Related (same bug class, separate repos)

- milo-os/milo — create rules (separate PR)
- milo-os/billing — billingaccountbinding (separate PR)
- Runbook: milo-os/activity#213 · Investigation: milo-os/activity#212

### DIFF
--- a/config/milo/activity/policies/backendtlspolicy-policy.yaml
+++ b/config/milo/activity/policies/backendtlspolicy-policy.yaml
@@ -16,12 +16,12 @@ spec:
   auditRules:
     # BackendTLSPolicy creation with spec available
     - name: create
-      match: "!audit.user.username.startsWith('system:') && audit.verb == 'create' && has(audit.requestObject.spec)"
+      match: "!audit.user.username.startsWith('system:') && audit.verb == 'create' && has(audit.requestObject.spec) && audit.responseStatus.code >= 200 && audit.responseStatus.code < 300"
       summary: "{{ actor }} created backend TLS policy {{ link(audit.responseObject.metadata.name, audit.objectRef) }}"
 
     # BackendTLSPolicy creation fallback (no spec)
     - name: create-fallback
-      match: "!audit.user.username.startsWith('system:') && audit.verb == 'create'"
+      match: "!audit.user.username.startsWith('system:') && audit.verb == 'create' && audit.responseStatus.code >= 200 && audit.responseStatus.code < 300"
       summary: "{{ actor }} created {{ link('a backend TLS policy', audit.objectRef) }}"
 
     # BackendTLSPolicy deletion with responseObject.spec available (response contains the deleted resource)
--- a/config/milo/activity/policies/backendtlspolicy-policy.yaml
+++ b/config/milo/activity/policies/backendtlspolicy-policy.yaml
@@ -14,10 +14,13 @@ spec:
     kind: BackendTLSPolicy
 
   auditRules:
-    # BackendTLSPolicy creation with spec available
+    # BackendTLSPolicy creation with spec available.
+    # Rejected creates return a Status responseObject (no metadata.name); for
+    # generateName creates objectRef.name is also empty, so fall back to
+    # responseObject.details.name. has() is guarded per level.
     - name: create
       match: "!audit.user.username.startsWith('system:') && audit.verb == 'create' && has(audit.requestObject.spec)"
-      summary: "{{ actor }} created backend TLS policy {{ has(audit.responseObject.metadata.name) ? link(audit.responseObject.metadata.name, audit.objectRef) : link(audit.objectRef.name, audit.objectRef) }}"
+      summary: "{{ actor }} created backend TLS policy {{ has(audit.responseObject.metadata.name) ? link(audit.responseObject.metadata.name, audit.objectRef) : has(audit.objectRef.name) ? link(audit.objectRef.name, audit.objectRef) : (has(audit.responseObject.details) && has(audit.responseObject.details.name)) ? link(audit.responseObject.details.name, audit.objectRef) : link('a backend TLS policy', audit.objectRef) }}"
 
     # BackendTLSPolicy creation fallback (no spec)
     - name: create-fallback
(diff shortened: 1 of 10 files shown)


scotwells:
Interesting, does the log capture what the full audit log payload is? I'm curious why the metadata name isn't available in the response object of the create operation.

ecv:
Successful creates have it, only failures don't. 

Quoth the raven, 

> Our `create` rule matches on `has(audit.requestObject.spec)` (the *request*, always valid), so a reject still matches the rule, then the summary derefs `responseObject.metadata.name` on a Status object → `no such key: name` → DLQ.

> The line in the description is the processor's CEL eval error, not the raw audit payload. The full `responseObject` is only present in the audit event if the apiserver audit level is `RequestResponse` (at `Metadata`/`Request` it isn't logged at all).

But `audit.objectRef.name`, which is populated on both success and reject, so. Safer.

ecv:
Yes — the processor logs the full audit event. `activityprocessor/processor.go:945` ("Failed to evaluate policy") emits the whole thing as `eventJSON`. Pulled a live one from staging (a `networking.datumapis.com-connector` create currently re-failing the DLQ retry):

```json
{
  "auditID": "5c93f394-87b7-4108-912a-920facc12774",
  "level": "RequestResponse",
  "verb": "create",
  "objectRef": {
    "apiGroup": "networking.datumapis.com",
    "apiVersion": "v1alpha1",
    "namespace": "default",
    "resource": "connectors"
  },
  "requestObject": {
    "kind": "Connector",
    "metadata": { "generateName": "datum-connect-" },
    "spec": { "connectorClassName": "datum-connect" }
  },
  "responseObject": {
    "apiVersion": "v1",
    "kind": "Status",
    "code": 403,
    "status": "Failure",
    "reason": "Forbidden",
    "message": "connectors.networking.datumapis.com \"datum-connect-2qjq6\" is forbidden: Your request took too long to be checked against your quota. Please try again in a moment — if this keeps happening, contact support.",
    "metadata": {},
    "details": { "group": "networking.datumapis.com", "kind": "connectors", "name": "datum-connect-2qjq6" }
  },
  "responseStatus": { "code": 403, "status": "Failure", "reason": "Forbidden", "details": { "name": "datum-connect-2qjq6" } }
}
```

This confirms the root cause concretely: the create was **rejected** (403, quota check timed out), so `responseObject` is a `metav1.Status` with `metadata: {}` — no `metadata.name`. Hence `no such key: name`.

**But this payload also shows the current fix is incomplete for `connector`.** Note `objectRef` has **no `name`**:

```json
"objectRef": { "apiGroup": "...", "apiVersion": "v1alpha1", "namespace": "default", "resource": "connectors" }
```

The client used `generateName: datum-connect-`, and the request was rejected at the quota check **before a name was assigned**, so `objectRef.name` is absent too. Our fallback —

```
has(audit.responseObject.metadata.name)
  ? link(audit.responseObject.metadata.name, audit.objectRef)
  : link(audit.objectRef.name, audit.objectRef)
```

— takes the else branch and derefs `audit.objectRef.name`, which raises `no such key: name` **again**. The processor logs confirm it: `networking.datumapis.com-connector` events are still re-failing the DLQ retry (retryCount climbing 0→10) on the fixed branch.

The name *is* present in the rejected event — at `responseObject.details.name` / `responseStatus.details.name` (`datum-connect-2qjq6`). So the fallback chain needs to go further for `generateName` resources, e.g.:

```
has(audit.responseObject.metadata.name) ? link(audit.responseObject.metadata.name, audit.objectRef)
  : has(audit.objectRef.name) ? link(audit.objectRef.name, audit.objectRef)
  : has(audit.responseObject.details.name) ? link(audit.responseObject.details.name, audit.objectRef)
  : link('a connector', audit.objectRef)
```

Gateway likely uses explicit names (so `objectRef.name` is populated and the current fix holds there), but any policy whose clients create via `generateName` — connector at minimum — needs the deeper fallback. I'll audit which of the 6 are generateName-based.

scotwells (approved):

Issue #231, opened 2026-06-29
fix: detachHTTPRoutes checks Status.Parents instead of Spec.ParentRefs during gateway finalization


ecv:
## Problem

\`finalizeGateway\` calls \`detachHTTPRoutes\` to find and delete downstream HTTPRoutes that reference the gateway being finalized. The function iterates **\`route.Status.Parents\`** (written by Envoy Gateway, can be stale or empty) instead of **\`route.Spec.ParentRefs\`** (written by NSO, always authoritative).

```go
// gateway_controller.go:1578 — detachHTTPRoutes
for _, parent := range route.Status.Parents {
    if string(parent.ParentRef.Name) == gateway.Name {
        continue
    }
    parents = append(parents, parent)
}
if len(parents) == 0 && deleteWhenNoParents {
    gatewayClient.Delete(ctx, &route)
}
```

**Why this is wrong**: \`Status.Parents\` is controlled by Envoy Gateway and reflects EG's reconciliation state, not what NSO wrote. It can diverge from \`Spec.ParentRefs\` when:

- EG hasn't yet processed the route at finalization time
- EG's status aggregation (Karmada → source) lags behind spec changes
- EG sets a controller-specific name in \`Status.Parents[].controllerName\` but the parent ref name matches — removing it correctly clears the entry, but the logic is fragile

In the observed incident, the finalization ran correctly enough to delete the HTTPRoute and GC the owned Service, but the orphaned HTTPRoute then re-appeared on edge clusters because Karmada's deletion propagation had a lag window with no upstream to drive re-cleanup. The HTTPRoute remained stuck in \`BackendNotFound\` for 37h+ with no auto-heal path.

## Fix

Change \`detachHTTPRoutes\` to check **\`Spec.ParentRefs\`** to identify whether a route references the gateway being finalized, and use that to determine whether the route should be deleted:

```go
// Correct: check Spec.ParentRefs — NSO writes these, they are authoritative
for _, ref := range route.Spec.ParentRefs {
    if ptr.Deref(ref.Group, gatewayv1.GroupName) == gatewayv1.GroupName &&
        ptr.Deref(ref.Kind, KindGateway) == KindGateway &&
        string(ref.Name) == gateway.Name {
        continue // this ref is to the gateway being deleted
    }
    remainingRefs = append(remainingRefs, ref)
}
if len(remainingRefs) == 0 && deleteWhenNoParents {
    gatewayClient.Delete(ctx, &route)
[...]

PR #234, opened 2026-06-29
feat: add orphaned downstream HTTPRoute GC controller

### LINKED ISSUE #233: feat: scan for and clean up orphaned downstream HTTPRoutes after gateway finalization
## Background

NSO v0.23.5 had a bug in `finalizeGateway` → `detachHTTPRoutes` ([#231](https://github.com/datum-cloud/network-services-operator/issues/231)): it checked `Status.Parents` instead of `Spec.ParentRefs` to decide which HTTPRoutes to delete when a Gateway was finalized. The fix shipped in v0.23.6.

However, the fix only prevents future occurrences. Any HTTPRoute orphaned by a past buggy finalization will never be cleaned up automatically — the upstream Gateway is already deleted, the finalizer has already run, and there is no reconciliation loop for a resource that no longer exists.

**Observed instance**: `HTTPRoute tunnel-858j6` orphaned on 6 alice edge clusters (ae-north-1, br-east-1, in-west-1, nl-west-1, us-east-2, us-west-1) for 37h+, `BackendNotFound`, because the upstream Gateway was deleted and finalization left the downstream HTTPRoute behind. Required manual Karmada deletion to resolve. Tracked in datum-cloud/infra#2901.

## Proposed solution

Add a periodic reconciliation loop (or a reconciler triggered on Karmada cluster membership events) that:

1. Lists all downstream HTTPRoutes across Karmada member clusters
2. For each HTTPRoute, checks whether a corresponding upstream Gateway still exists in the project cluster
3. If no upstream Gateway exists and the HTTPRoute has no other valid `Spec.ParentRefs`, deletes the orphaned HTTPRoute from Karmada

This is the same "orphan detection" pattern used by garbage collectors — walk the owned resources, verify the owner still exists, delete if not.

## Acceptance criteria

- Orphaned downstream HTTPRoutes (no corresponding upstream Gateway) are detected and deleted within one reconciliation interval
- The scan does not delete HTTPRoutes that legitimately have no upstream yet (e.g. during initial propagation lag) — consider a minimum age threshold
- Existing orphans from pre-v0.23.6 deployments are cleaned up after upgrade without manual intervention
- Covered by integration test: create Gateway → delete Gateway → verify downstream HTTPRoute is gone (even if `detachHTTPRoutes` fails to clean it up)

## Related

- [#231](https://github.com/datum-cloud/network-services-operator/issues/231) — root cause fix (`detachHTTPRoutes` Spec.ParentRefs)
- datum-cloud/infra#2901 — incident: tunnel-858j6 orphaned on 6 clusters, 37h+ firing
- datum-cloud/engineering#330 — tunnels outage incident

ecv:
## Summary

- Adds `OrphanedDownstreamHTTPRouteGCReconciler` that scans Karmada (downstream cluster) for HTTPRoutes that reference a Gateway which no longer exists on the upstream project cluster, and deletes them
- Watches both upstream HTTPRoutes and downstream Karmada HTTPRoutes so orphans from before NSO v0.23.6 are caught on startup via the initial informer list
- 5-minute minimum age threshold prevents deletion during normal Karmada propagation lag
- A route with multiple parentRefs is only deleted when all referenced upstream Gateways are gone

## Motivation

The v0.23.5 bug fixed in #231 (`detachHTTPRoutes` using `Status.Parents`) could leave downstream HTTPRoutes alive on Karmada even after the upstream Gateway was deleted. Since the finalizer already ran, those orphans survive the v0.23.6 rollout and need a separate scan-and-delete pass. They also cause `HTTPRouteStatusWarning/BackendNotFound` alerts that fire indefinitely with no auto-heal.

Incident: datum-cloud/infra#2901, datum-cloud/engineering#330

## Test plan

- [x] `TestOrphanedDownstreamHTTPRouteGC/orphaned_route_deleted_when_upstream_gateway_missing`
- [x] `TestOrphanedDownstreamHTTPRouteGC/healthy_route_not_deleted_when_upstream_gateway_exists`
- [x] `TestOrphanedDownstreamHTTPRouteGC/young_route_not_deleted;_requeue_returned`
- [x] `TestOrphanedDownstreamHTTPRouteGC/route_with_multiple_parentRefs_kept_if_any_gateway_still_exists`
- [x] `TestOrphanedDownstreamHTTPRouteGC/no_downstream_routes_is_a_no-op`
- [x] Full controller package test suite passes (`go test ./internal/controller/...`)
- [ ] Verify existing orphan `tunnel-858j6` is cleaned up after this ships (datum-cloud/infra#2901)

Closes #233

### DIFF
--- a/internal/cmd/manager/manager.go
+++ b/internal/cmd/manager/manager.go
@@ -84,528 +84,536 @@ type BuildInfo struct {
 	BuildDate    string
 }
 
+type managerConfig struct {
+	build                                BuildInfo
+	serverConfigFile                     string
+	probeAddr                            string
+	enableLeaderElection                 bool
+	leaderElectionNamespace              string
+	enableClusterSharding                bool
+	clusterShardingLeaseNamespace        string
+	clusterShardingLeasePrefix           string
+	clusterShardingPeerWeight            uint
+	singletonControllersLeaderElection   bool
+	singletonControllersLeaderElectionID string
+	opts                                 zap.Options
+}
+
 // NewCommand builds the "manager" subcommand, which runs the
 // network-services-operator controller manager.
 func NewCommand(build BuildInfo) *cobra.Command {
-	var enableLeaderElection bool
-	var leaderElectionNamespace string
-	var probeAddr string
-	var enableClusterSharding bool
-	var clusterShardingLeaseNamespace string
-	var clusterShardingLeasePrefix string
-	var clusterShardingPeerWeight uint
-	var singletonControllersLeaderElection bool
-	var singletonControllersLeaderElectionID string
-
-	var serverConfigFile string
+	cfg := managerConfig{
+		build: build,
+		opts:  zap.Options{Development: true},
+	}
 
 	fs := flag.NewFlagSet("manager", flag.ContinueOnError)
 
-	fs.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	fs.BoolVar(&enableLeaderElection, "leader-elect", false,
+	fs.StringVar(&cfg.probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	fs.BoolVar(&cfg.enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	fs.StringVar(&leaderElectionNamespace, "leader-elect-namespace", "", "The namespace to use for leader election.")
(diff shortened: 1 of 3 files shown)


scotwells:
Seems like this is a bandaid for a garage collection issue? I wouldn't want to operationalize a bandaid.

ecv:
Mmm, we could delete the orphans from Karmada, but I'm preferring not to issue one-off administrative actions. 

I suppose we could develop a scripted admin process for deleting stuff from Karmada in the event we have to do so. But, shouldn't the component take care of them?

The underlying orphan-creator has (probably) [been fixed](https://github.com/datum-cloud/network-services-operator/issues/231).

ecv:
Doesn't seem like the most sane implementation.

ecv:
Closing as won't-do, sense of group is we do this as a one-off.

PR #237, opened 2026-07-01
test: migrate httproute-accepted e2e from infra

ecv:
## What

Adds an `httproute-accepted` end-to-end test to NSO's local kind harness at `test/e2e/httproute-accepted/`. It provisions a `Gateway` and an `HTTPRoute` and asserts the route reaches `Accepted`, so downstream traffic programming can proceed.

## Why

Per `datum-cloud/infra#3006`, single-service API-contract tests are moving out of the infra live-env Chainsaw suite into their owning repo's local kind harness, run pre-merge. The `gateway.networking.k8s.io` HTTPRoute contract is owned by NSO, so it belongs here. Related: `datum-cloud/infra#3005`.

## Approach

The test mirrors the `gateway` sibling's proven setup on the prod-fidelity harness:

- a self-signed CA on the downstream cluster (`nso-infra`),
- an upstream `GatewayClass`,
- a verified `Domain` (`e2e.env.datum.net`),
- a `Gateway` with HTTP + HTTPS listeners that reaches `Accepted`,
- an `HTTPRoute` + backend `EndpointSlice`, then an assert that the route is `Accepted`.

Two details worth calling out, both learned from CI:

> [!NOTE]
> Route acceptance is asserted on the per-parent condition (`status.parents[].conditions[]`), where the Gateway API surfaces it — the HTTPRoute never populates a top-level `Accepted` condition, so a `kubectl wait --for=condition=Accepted` can't observe it.

> [!NOTE]
> The `conditions` list asserts both `Accepted` and `ResolvedRefs`. Chainsaw matches a bare array positionally and by length, so an expected list carrying only `Accepted` never matches the route's two-entry status.

## Test plan

- `chainsaw lint` passes.
- E2E Tests green end-to-end: `httproute-accepted` passes in ~7s alongside the rest of the suite. This is verified against a running harness, not just linted.

### DIFF
--- a/test/e2e/httproute-accepted/chainsaw-test.yaml
+++ b/test/e2e/httproute-accepted/chainsaw-test.yaml
@@ -1,0 +1,215 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: httproute-accepted
+spec:
+  bindings:
+    - name: clusterIssuerName
+      value: (join('-', ['e2e', $namespace]))
+    - name: gatewayClassName
+      value: (join('-', ['e2e', $namespace]))
+  cluster: nso-standard
+  steps:
+    - name: Create CA
+      try:
+        - create:
+            cluster: nso-infra
+            resource:
+              apiVersion: cert-manager.io/v1
+              kind: ClusterIssuer
+              metadata:
+                name: (join('-', [$clusterIssuerName, 'issuer']))
+              spec:
+                selfSigned: {}
+
+        - create:
+            cluster: nso-infra
+            resource:
+              apiVersion: cert-manager.io/v1
+              kind: Certificate
+              metadata:
+                name: (join('-', [$clusterIssuerName, 'ca']))
+                namespace: cert-manager
+              spec:
+                isCA: true
+                commonName: (join('-', [$clusterIssuerName, 'ca']))
+                secretName: ($clusterIssuerName)
+                privateKey:
+                  algorithm: ECDSA
+                  size: 256
+                issuerRef:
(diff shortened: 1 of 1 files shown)


ecv:
## ⚠️ Action required before this leaves draft

Two things to verify against a running `nso-standard`/`nso-infra` kind setup:

1. **Is the preamble sufficient for `Accepted`?** This test mirrors the `gateway` sibling's
   cert-manager CA + downstream EnvoyProxy/GatewayClass + Domain preamble, gives the listener
   a hostname (required post-#219/#220), and seeds a backend `EndpointSlice`. Confirm the
   Gateway actually programs and reaches `Accepted` locally.
2. **HTTPRoute condition shape.** Confirm the HTTPRoute exposes a top-level `Accepted`
   condition vs only `status.parents[].conditions[]` — the `wait` may need to target the
   parent-scoped condition instead.

Not run end-to-end yet (only `chainsaw lint`). Tracks datum-cloud/infra#3006.

ecv:
Verified end-to-end now — the draft caveats above are resolved.

The branch was well behind `main`, so I rebased it onto the current prod-fidelity harness and reworked the test against it. What actually needed fixing, in order:

1. **HTTPRoute acceptance lives on the parent, not the top level.** The route is admitted under `status.parents[].conditions[]`; there's no top-level `Accepted` condition, so the original `kubectl wait --for=condition=Accepted` could never observe it and always timed out.
2. **The old manual downstream setup is gone.** The prod-fidelity harness now pre-installs Envoy Gateway (in `datum-downstream-gateway`) and the downstream `GatewayClass`, so the test no longer copies a CA into `envoy-gateway-system` or hand-rolls an `EnvoyProxy`/`GatewayClass`.
3. **Chainsaw array matching is positional and length-sensitive.** Asserting only `Accepted` never matched the route's two-entry status, so the assert now carries both `Accepted` and `ResolvedRefs`, mirroring the `gateway` sibling.

E2E Tests is green: `httproute-accepted` passes in ~7s and the rest of the suite is unaffected. Marking ready for review.

scotwells (approved):

PR #238, opened 2026-07-03
feat: add forwardIDToken to oidc config in SecurityPolicy

savme:
The Milo control plane installs EG CRDs from v1.5.1, while edge clusters run v1.8.1. The `forwardIDToken` OIDC field was added after v1.5 release, so it cannot be set on a SecurityPolicy even though the downstream cluster would accept it.

A full version bump from v1.5.1 -> v1.8.1 would expose 50+ new fields - several (TLS cipher control, admission control, dynamic Wasm modules, IP geo filtering) need to be validated before being made available on the control plane. This PR adds just the field that is required by the datum wiki deployment.

Key changes:
- Add a kustomize JSON patch that injects `forwardIDToken` into the SecurityPolicy OIDC schema, keeping v1.5.1 as the base
- Include the companion CEL rule from v1.8.1 that prevents setting `forwardAccessToken: true` alongside `forwardIDToken.header: Authorization` — without it, that combination would be silently accepted upstream then rejected by the downstream v1.8.1 CRD

### DIFF
--- a/config/crd/gateway/kustomization.yaml
+++ b/config/crd/gateway/kustomization.yaml
@@ -20,6 +20,13 @@ patches:
       kind: CustomResourceDefinition
       metadata:
         name: referencegrants.gateway.networking.k8s.io
+  # Backports fields from EG v1.8.1. Re-evaluate this when upgrading the gateway CRDs.
+  - path: patch-securitypolicy-forward-id-token.yaml
+    target:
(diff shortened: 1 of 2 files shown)


savme:
Closing because I learned that support for this in v1.8 is a bit funky. Thanks for saving me the trouble of figuring this out post-deploy, @scotwells! 🙏

PR #241, opened 2026-07-08
fix(activity-policies): gate create rules on 2xx outcome

ecv:
## Summary

Failed create requests (e.g. a `403`) return a `metav1.Status` object as the audit `responseObject`, not the created resource. ActivityPolicy summary templates that dereference `audit.responseObject.metadata.name` therefore threw on those payloads, routing the events to the DLQ (`DLQSlowLeak`). Ungated create rules also emitted false "created" activities for attempts that actually failed.

This gates **every** `create` rule on the request outcome by appending `audit.responseStatus.code >= 200 && audit.responseStatus.code < 300` to the `match` expression, so failed attempts produce no activity. This matches the pattern milo standardized on in commit `2b82751`.

## Changes

All 8 policies in `config/milo/activity/policies/` — both create rules (the primary create + the no-spec `create-fallback`) are gated in each:

- `domain-policy.yaml`
- `httpproxy-policy.yaml`
- `httproute-policy.yaml`
- `gateway-policy.yaml`
- `connector-policy.yaml` — the policy named in the issue; its `create` summary derefs `audit.responseObject.metadata.name`, which was the DLQ trigger
- `connectoradvertisement-policy.yaml`
- `trafficprotectionpolicy-policy.yaml`
- `backendtlspolicy-policy.yaml`

16 create-rule `match` expressions updated total. Create rules with existing trailing `has(...)`/`size(...)` clauses keep those clauses; the outcome guard is appended after them. Delete/update/patch rules are untouched.

Policies ship as raw YAML referenced directly by `kustomization.yaml`; no generate/bundle step applies.

Tracks milo-os/activity#215.

### DIFF
--- a/config/milo/activity/policies/backendtlspolicy-policy.yaml
+++ b/config/milo/activity/policies/backendtlspolicy-policy.yaml
@@ -16,12 +16,12 @@ spec:
   auditRules:
     # BackendTLSPolicy creation with spec available
     - name: create
-      match: "!audit.user.username.startsWith('system:') && audit.verb == 'create' && has(audit.requestObject.spec)"
+      match: "!audit.user.username.startsWith('system:') && audit.verb == 'create' && has(audit.requestObject.spec) && audit.responseStatus.code >= 200 && audit.responseStatus.code < 300"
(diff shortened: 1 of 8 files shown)


ecv:
this holds until e2e tests are all stable.

Issue #243, opened 2026-07-10
extension-server: branded error page injected as Envoy SubstitutionFormatString → unescaped `%` rejects the listener (breaks TPP Enforce)


ecv:
## Summary

The extension-server injects the branded error page into Envoy's `local_reply_config` as a **`SubstitutionFormatString` / `TextFormatSource`**. Envoy parses that string as a format string, so every literal `%` in the HTML is treated as a command operator. The embedded default page contains unescaped `%` characters, so Envoy **rejects the entire listener** — and with the downstream Envoy Gateway running `failOpen: false`, the rejected xDS update is dropped fleet-wide, leaving no listener. Result: 100% traffic failure on any `TrafficProtectionPolicy` in `Enforce` mode that falls back to the embedded page.

Discovered while reproducing #242 on a full local kind WAF stack.

## Evidence

Envoy on the merged downstream gateway:

```
[warning][config] delta config for type.googleapis.com/envoy.config.listener.v3.Listener
  rejected: Error adding/updating listener(s) tcp-80: Incorrect configuration: <!DOCTYPE html>
  ...
  . Couldn't find valid command at position 330
```

`/listeners` then shows only the admin/stats listeners (no `tcp-80`), and clients get `curl: (52) Empty reply from server`. The backend is healthy (direct hit returns 200). Removing the TPP restores the listener.

Byte 330 of `internal/extensionserver/assets/error-5xx-default.html` is the `%` in:

```css
html, body { height: 100%; margin: 0; }
```

The embedded page has 12 single `%` and zero `%%` escapes.

## Root cause

`internal/extensionserver/mutate/localreply.go` (`buildLocalReplyConfig`):

```go
BodyFormatOverride: &corev3.SubstitutionFormatString{
    ContentType: cfg.ContentType,
    Format: &corev3.SubstitutionFormatString_TextFormatSource{
        TextFormatSource: &corev3.DataSource{
            Specifier: &corev3.DataSource_InlineString{InlineString: cfg.BodyHTML},
        },
    },
},
```

`%RESPONSE_CODE%` is the intended (valid) substitution, but any bare `%` in the body is not. `error-5xx-default.html` does not escape its literal `%`. The infra-mounted override page (`apps/envoy-error-pages/.../error-5xx.html`) *does* escape them (`width: 100%%`), which is why prod hasn't been taken down by this specific path — but any gateway without the `envoy-error-pages` ConfigMap falls back to the malformed embedded page.

## Fix

[...]

PR #245, opened 2026-07-10
test: e2e for TrafficProtectionPolicy Enforce serving traffic (#243)

ecv:
## What

Chainsaw e2e regression for #243. Deploys a backend, a Gateway on the WAF `datum-downstream-gateway` class with an HTTPRoute, and a `TrafficProtectionPolicy` in `Enforce` mode, then probes the downstream Envoy over the wire and asserts a benign `GET` returns **200** — not the empty/5xx reply the unescaped error-page body produces.

`Gateway.status.Programmed` alone is insufficient: Envoy Gateway reports it from its own translation, independent of Envoy's xDS ACK. So the test asserts on an actual HTTP response.

## Precondition

Requires the WAF data plane on the downstream cluster (extension-server + the extensionManager Envoy Gateway registered on the `datum-downstream-gateway` class + the Coraza filter). Stock `make prepare-infra-cluster` does not install this; environments without it skip via the unmet `Programmed` assertion.

## Proof (this exact test, same fresh-fleet harness, only the operator image differs)

- ❌ **Negative — pre-fix image:** TPP Enforce → `HTTP 000` empty reply, test **FAILS** → https://gist.github.com/ecv/e47733b657475887c4c29cced339512a
- ✅ **Positive — fixed image (#244):** TPP Enforce → `HTTP 200`, test **PASSES** → https://gist.github.com/ecv/4c601d3d96fd5b317c5f3b814bbdd510

The fix is in #244.

Refs #243.

### DIFF
--- a/test/e2e/trafficprotectionpolicy-enforce/chainsaw-test.yaml
+++ b/test/e2e/trafficprotectionpolicy-enforce/chainsaw-test.yaml
@@ -22,9 +22,6 @@ spec:
     metadata:
       labels:
         meta.datumapis.com/upstream-cluster-name: e2e
-  bindings:
-    - name: hostname
-      value: (join('.', [$namespace, 'e2e.test']))
   steps:
     - name: Deploy a backend
       try:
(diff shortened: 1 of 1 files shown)


ecv:
oh ok, the red test in this is because the fixes arent in this branch. it proves the negative right now. whew

kevwilliams (approved):
I like this

PR #255, opened 2026-07-12
fix: retrigger EG translation on TPP spec changes

### LINKED ISSUE #254: TrafficProtectionPolicy mode/spec changes don't propagate to edge (no EG re-translation trigger for TPP)
## Summary

A `TrafficProtectionPolicy` (TPP) **mode/spec change does not reliably propagate to the edge data plane.** The new value lands in the extension server's cache immediately, but Envoy Gateway is never told to re-translate, so the edge keeps serving the **stale** WAF config until an unrelated event incidentally forces a re-translation — anywhere from seconds to >10 min later, with no upper bound.

## Impact

- **Product:** flipping a TPP `Observe ↔ Enforce` (or any WAF-affecting spec change) may not take effect on live traffic for minutes, with no signal that the edge is stale. Enforce can keep blocking after a user disables it, or fail to block after a user enables it.
- Breaks the `waf-configuration` construct e2e test, which flips a TPP `Observe → Enforce` and asserts the block within ~8 min. See datum-cloud/infra#3402 for the full live repro.

## Root cause

Coraza's `SecRuleEngine On`/`DetectionOnly` — the Enforce/Observe distinction — is injected into the edge Envoy as inline xDS filter config via Envoy Gateway's `PostTranslateModify` extension hook, computed at hook time from the extension server's local TPP informer cache (`internal/extensionserver/cache/index.go`, `internal/extensionserver/mutate/tpp.go`, `internal/extensionserver/server/server.go`).

Envoy Gateway only re-runs translation (and thus the hook) when a resource **it natively watches** changes. A `TrafficProtectionPolicy` is not such a resource. There is a dedicated retrigger controller for exactly this problem — but it watches **`Connector` only, not TPP**:

- `internal/extensionserver/retrigger/retrigger.go` — `Reconciler` is wired `For(&Connector{})`.
- `internal/extensionserver/cmd/run.go` — the edge re-translation controller registers only the Connector `retrigger.Reconciler`; there is no TPP equivalent.

So a TPP mode flip updates the extension cache promptly, but nothing forces EG to re-translate → the data plane serves the stale program until unrelated namespace churn or EG's periodic resync happens to trigger a re-translation. That incidental timing is the observed variable, heavy-tailed delay.

## Why it was masked until recently

The `TrafficProtectionPolicyReconciler` formerly wrote `EnvoyPatchPolicy` objects, which **are** EG-watched, so a mode flip implicitly retriggered translation. EPP emission is now gated behind `IsEPPEmissionEnabled()` (`internal/config/config.go`) and turned **off** in the datum-cloud/infra prod and staging deployments (`eppEmissionEnabled: false`). With EPP off, a mode flip touches no EG-watched resource. The gap has been latent since the extension-server rollout (infra `408e30661`, 2026-06-17); the test passed before only when a post-flip reconcile happened to land in the fast tail.

## Fix direction

Add a TPP retrigger controller symmetric to the Connector one in `internal/extensionserver/retrigger/`: watch `TrafficProtectionPolicy`, and on a spec/mode change touch each owning Gateway's trigger annotation so EG re-translates against the fresh cache. Extend to any other WAF-affecting spec the extension server reads from cache.

## Decisive confirmation

Flip a TPP mode, then `kubectl annotate` the owning Gateway (mimicking the Connector retrigger) — the change should appear on the data plane within seconds.

## Refs

- datum-cloud/infra#3402 (live repro, distribution, impact)
- Source: `internal/extensionserver/{retrigger/retrigger.go,cmd/run.go,cache/index.go,mutate/tpp.go,server/server.go}`, `internal/controller/trafficprotectionpolicy_controller.go`, `internal/config/config.go`

### LINKED ISSUE #256: TPP retrigger: handle deletion and targetRef removal
## Summary

Follow-up to #254 / #255. The TPP retrigger controller (`internal/extensionserver/retrigger/tpp.go`) forces Envoy Gateway to re-translate on a `TrafficProtectionPolicy` **create or spec change**, so mode/spec flips reach the edge promptly. Two related cases are **not** covered and can leave stale WAF config on the data plane:

### 1. TPP deletion

When a TPP is deleted, its governed Gateways are never nudged, so the edge keeps serving the last WAF program (e.g. an `Enforce` policy keeps **blocking** after the policy is gone) until an unrelated EG-watched resource incidentally re-translates.

Root cause: the controller's predicate drops deletes, and standard `Reconcile` receives only the object key — not the last-known object — so on delete there are no `targetRefs` to resolve owning Gateways from.

### 2. targetRef removal

On a targetRef edit, only the **current** targets are touched. A Gateway dropped from `spec.targetRefs` is not nudged, so it keeps the old WAF config even though the policy no longer governs it.

## Impact

Same class of failure as #254 (stale WAF on the edge, no signal), narrower trigger: disabling protection by **deleting** a TPP (or **removing a target**) may not take effect for minutes-to-never in a quiet namespace. Enforce can keep blocking after a user removes the policy.

## Fix direction

Track enough state to know the **previous** owning Gateway set, then touch the union of old and new targets:

- Use a mapping event handler (`Watches` + `handler.EnqueueRequestsFromMapFunc`) that captures the deleted/old object's `targetRefs`, or
- Record last-applied targets (annotation on the TPP, or an in-memory map keyed by TPP) and diff on each reconcile, or
- A finalizer that fires the retrigger on deletion before the object is removed (heavier; TPPs are replicated edge objects).

On delete/removal, touch each previously-owning Gateway so EG re-translates and drops the now-orphaned WAF config.

## Refs

- #254 (root cause), #255 (create/spec-change fix)
- Source: `internal/extensionserver/retrigger/tpp.go`

ecv:
## Summary

Fixes #254 and #256. A `TrafficProtectionPolicy` mode/spec change updates the extension server's local cache promptly, but Envoy Gateway only re-runs translation — and therefore the WAF-injecting extension hook — when a resource **it natively watches** changes. A TPP is not such a resource, so the edge keeps serving the **stale** WAF program until some unrelated EG-watched resource incidentally forces a re-translation. In a quiet namespace (exactly the CI test's shape) that never happens, so a mode flip can stay unenforced (or stuck enforcing) indefinitely.

There is already a dedicated retrigger controller for this class of problem, but it watches **`Connector` only**. This adds the symmetric TPP arm and makes it complete: create/spec-change, deletion, and targetRef removal all reach the edge.

## Change

- **`internal/extensionserver/retrigger/tpp.go`** — new `TPPReconciler`, symmetric to the existing `Connector` `Reconciler`. Watches `TrafficProtectionPolicy`; on a spec change it touches a per-TPP trigger annotation on each owning Gateway so EG re-translates against the fresh cache. On deletion or targetRef removal it **clears** the annotation on the Gateways that dropped out of the target set, so EG re-translates against the now-empty cache and drops the orphaned WAF program.
- **`internal/extensionserver/cmd/run.go`** — registers the new controller alongside the Connector arm.
- **`internal/extensionserver/retrigger/tpp_test.go`** — unit tests (stamp, mode-flip, missing Gateway, multi-target, per-TPP key, predicate, deletion, targetRef removal).

### Design notes

- **Predicate:** admits creates, generation bumps (any spec change — mode, targetRefs, ruleSets, sampling), and **deletes**. The cache reads spec fields only, so status/metadata churn is ignored.
- **Per-TPP annotation key** (`networking.datumapis.com/tpp-generation-<name>`): two TPPs targeting the same Gateway each own their own slot, so they never clobber each other's value and flip-flop it on every resync.
- **Merge patch, no preceding Get:** an unchanged generation (or clearing an already-absent annotation) is a server-side no-op (no resourceVersion bump, no EG event), so it is idempotent and never triggers a spurious re-translation.
- **Previous-target tracking (delete / targetRef removal):** standard `Reconcile` receives only the object key on delete, not the last-known object, so there are no `targetRefs` to resolve owning Gateways from; a targetRef edit only exposes the current targets. The controller records each TPP's owning Gateway set in an in-memory map keyed by namespaced name, populated by the create event every live TPP fires on startup (so a restart re-learns the set before any later edit or delete). On delete it clears the recorded set; on a targetRef edit it clears the members that dropped out. On a failed touch of a current target the previous set is retained so dropped Gateways are retried on requeue rather than forgotten.
- **TargetRef → Gateway mapping is name-derived and local:** Gateway targetRefs name the Gateway directly; HTTPRoute targetRefs share the HTTPProxy name, which is also the Gateway name (`httpproxy_controller.go` names Gateway, HTTPRoute, and HTTPProxy identically).

## Decisive confirmation (per the issues)

- Flip a TPP mode in a quiet namespace; the change lands on the data plane within seconds rather than never.
- Delete an `Enforce` TPP (or remove a target); the edge stops blocking within seconds rather than serving the stale program until incidental churn.
- Re-run the `waf-configuration` construct e2e (datum-cloud/infra#3402).

## Known limitations

- A TPP created **and** deleted while this controller is down leaves no event on restart, so its Gateways are not re-translated (stale WAF persists). Same restart-window caveat as the Connector arm; acceptable given the in-memory tracking.

Closes #254
Closes #256

### DIFF
--- a/internal/extensionserver/cmd/run.go
+++ b/internal/extensionserver/cmd/run.go
@@ -246,11 +246,17 @@ func run(o options) {
 		os.Exit(1)
 	}
 
-	// --- Edge re-translation controller ---
+	// --- Edge re-translation controllers ---
 	// Makes Envoy Gateway re-translate when a Connector's liveness changes, so
 	// the hook re-runs against fresh liveness. See the retrigger package doc.
 	if err := (&retrigger.Reconciler{Client: mgr.GetClient()}).SetupWithManager(mgr); err != nil {
-		log.Error("set up gateway re-translation controller", "err", err)
+		log.Error("set up connector re-translation controller", "err", err)
+		os.Exit(1)
+	}
+	// Symmetric arm for TrafficProtectionPolicy: a TPP is not EG-watched, so a
+	// mode/spec flip lands in the cache but never re-translates on its own.
+	if err := (&retrigger.TPPReconciler{Client: mgr.GetClient()}).SetupWithManager(mgr); err != nil {
+		log.Error("set up TPP re-translation controller", "err", err)
 		os.Exit(1)
 	}
 
--- a/internal/extensionserver/retrigger/tpp.go
+++ b/internal/extensionserver/retrigger/tpp.go
@@ -1,0 +1,149 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+// This file adds the TrafficProtectionPolicy arm of the edge re-translation
+// controller. It is symmetric to the Connector Reconciler in retrigger.go:
+// Envoy Gateway only re-runs translation — and so the extension hook — when a
+// resource it natively watches changes. A TrafficProtectionPolicy is not such a
+// resource, so a mode/spec flip lands promptly in the extension server's local
+// cache but the data plane keeps serving the stale WAF program until some
+// unrelated EG-watched resource incidentally forces a re-translation.
+//
+// This controller watches TrafficProtectionPolicies and, on a spec change (which
+// is exactly what the cache reads: mode, targetRefs, ruleSets, sampling), touches
+// a trigger annotation on each owning Gateway. Because the touch happens after
+// the new spec is already in the shared cache, EG re-translates against fresh
+// data and the mode change reaches the edge in seconds.
+
+package retrigger
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	networkingv1alpha "go.datum.net/network-services-operator/api/v1alpha"
+)
+
+// tppTriggerAnnotationPrefix is the annotation key prefix patched onto a Gateway
+// to make Envoy Gateway re-translate after a TPP spec change. The owning TPP's
+// name is appended so each policy owns its own annotation slot: two TPPs
+// targeting the same Gateway never clobber each other's value and flip-flop it
+// on every resync.
+const tppTriggerAnnotationPrefix = "networking.datumapis.com/tpp-generation-"
+
+// TPPReconciler touches the trigger annotation on every Gateway a
+// TrafficProtectionPolicy targets whenever the policy's spec changes, forcing EG
+// to re-translate and re-run the extension hook against the fresh cache.
+type TPPReconciler struct {
+	Client client.Client
+}
+
+// tppTriggerAnnotationKey returns the per-TPP trigger annotation key.
+func tppTriggerAnnotationKey(tppName string) string {
+	return tppTriggerAnnotationPrefix + tppName
+}
+
+// Reconcile stamps the TPP's observed generation onto the trigger annotation of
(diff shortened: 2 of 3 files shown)


ecv:
tests broke on main, not here; fixing there

yahyafakhroji (approved):

Issue #250, opened 2026-07-10
TrafficProtectionPolicy: blocking paranoia level > detection triggers CRS 901500 → 100% of traffic gets HTTP 500


ecv:
## Summary

A `TrafficProtectionPolicy` (OWASP CRS) whose **blocking** paranoia level is greater than its **detection** paranoia level causes OWASP CRS rule **901500** to `deny` **every request with HTTP 500** — before any attack evaluation. The two fields are validated independently (each `1..4`, default `1`) with no cross-field constraint, so the illegal combination is admissible. It is easy to hit by accident: `blocking` defaults to `1`, so setting `blocking: 2` while leaving `detection` unset (default `1`) silently 500s 100% of traffic to the protected route.

This is live in prod: route `pcep-lol-hnqr6o` served **~120,730 HTTP 500s over 7 days** (effectively all its traffic) from this; `kev1n-proxy` and `kev-httpbin` also hit it at lower volume.

## Steps to reproduce

1. Attach a `TrafficProtectionPolicy` in `Enforce` mode to an `HTTPProxy`/route, with an `OWASPCoreRuleSet` rule set and:
   ```yaml
   paranoiaLevels:
     blocking: 2
     detection: 1   # or simply omit detection — it defaults to 1
   ```
2. Send **any** request to the protected route — a plain benign `GET /`, no attack payload.
3. Observe the response.

## Expected behavior

Either:
- the policy is **rejected at admission** because `detection < blocking` is an illegal CRS configuration (the field docs already state detection is meant to be the *higher* of the two), or
- `detection` is defaulted/clamped to `>= blocking`,

so that a benign request is evaluated normally (→ `200`, or a legitimate WAF verdict) rather than every request failing.

## Observed behavior

**Every** request to the route returns **HTTP 500** (empty body), regardless of content. CRS rule 901500 fires at phase 1:

```
id:901500, phase:1, deny, status:500,
msg:'Detection paranoia level configured is lower than the paranoia level itself. This is illegal. Blocking request. Aborting'
```

### Prod evidence (7d, prod VictoriaMetrics)

- `coraza_envoy_filter_request_events_total{trafficprotectionpolicy_name="pcep-lol-hnqr6o"}` → rule `901500`, action `deny`, ~603,638 events.
[...]

PR #258, opened 2026-07-12
fix: default TPP paranoiaLevels so CRD applies

### LINKED ISSUE #257: TrafficProtectionPolicy CRD rejected: paranoiaLevels default {} violates detection>=blocking CEL rule
## Summary

The `trafficprotectionpolicies.networking.datumapis.com` CRD is **rejected at apply time**, breaking `make deploy` / the e2e "Prepare e2e" step on every branch:

```
The CustomResourceDefinition "trafficprotectionpolicies.networking.datumapis.com" is invalid:
spec...properties[ruleSets].items.properties[owaspCoreRuleSet].properties[paranoiaLevels].default:
Invalid value: "object": no such key: detection
evaluating rule: detection paranoia level must be greater than or equal to blocking paranoia level
```

## Root cause

Regression from #251 (fix for #250). That change added a struct-level CEL rule on `ParanoiaLevels`:

```go
// +kubebuilder:validation:XValidation:message="detection paranoia level must be greater than or equal to blocking paranoia level",rule="self.detection >= self.blocking"
type ParanoiaLevels struct { ... }
```

but the field keeps its empty-object default:

```go
// +kubebuilder:default={}
ParanoiaLevels ParanoiaLevels `json:"paranoiaLevels,omitempty"`
```

The apiserver validates a schema's structural **default** against that node's `x-kubernetes-validations`. The default `{}` has no `detection`/`blocking` keys (per-field defaults are not injected before the parent default is CEL-checked), so `self.detection` resolves to *no such key* and CRD registration fails.

## Impact

- `kubectl apply` of the CRD fails, so **any** pipeline that deploys it — including every PR's e2e — goes red regardless of the PR's own changes (e.g. #255).

## Fix

Give the field a default that satisfies the rule:

```go
// +kubebuilder:default={blocking:1,detection:1}
ParanoiaLevels ParanoiaLevels `json:"paranoiaLevels,omitempty"`
```

`1 >= 1` passes; per-field defaults are unchanged; #251's strict rule still rejects inverted user-supplied values. Then regenerate the CRD (`make manifests`).

## Refs

- #250, #251 (source of the regression)
- `api/v1alpha/trafficprotectionpolicy_types.go`

ecv:
Fixes #257.

## Problem

The `trafficprotectionpolicies` CRD is rejected at apply time, so `kubectl apply` of the CRD fails and every branch's e2e "Prepare e2e" step goes red regardless of its own changes (e.g. #255):

```
The CustomResourceDefinition "trafficprotectionpolicies.networking.datumapis.com" is invalid:
spec...properties[ruleSets].items.properties[owaspCoreRuleSet].properties[paranoiaLevels].default:
Invalid value: "object": no such key: detection
evaluating rule: detection paranoia level must be greater than or equal to blocking paranoia level
```

## Root cause

Regression from #251 (fix for #250). It added a struct CEL rule on `ParanoiaLevels` (`self.detection >= self.blocking`) but the field kept `+kubebuilder:default={}`. The apiserver validates a schema node's structural default against that node's `x-kubernetes-validations`; the default `{}` has no `detection`/`blocking` keys (per-field defaults are not injected before the parent default is CEL-checked), so `self.detection` is *no such key* and CRD registration fails.

## Fix

Set the field default to `{blocking:1,detection:1}` so it satisfies `1 >= 1`. Per-field defaults are unchanged and #251's rule still rejects inverted user-supplied values. CRD manifest regenerated via `make manifests`.

## Regression guard

Adds an **envtest** suite (`test/crd/`) that installs the generated CRDs against a real apiserver and creates a defaulted TPP. The whole suite uses the fake client, which never sees schema/defaults/CEL — that is exactly why #251 slipped through. envtest reproduces the apiserver's default-vs-CEL validation, so an invalid default fails at CRD install. Verified: reverting the default to `{}` fails the suite with the exact error above.

- Installs OK + a policy with only its required targetRef defaults `paranoiaLevels` to `{blocking:1,detection:1}`.
- An inverted user-supplied pair (`blocking:3,detection:1`) is still rejected, so the default fix did not loosen #251's rule.
- Skips when `KUBEBUILDER_ASSETS` is unset; runs under `make test`.

### DIFF
--- a/api/v1alpha/trafficprotectionpolicy_types.go
+++ b/api/v1alpha/trafficprotectionpolicy_types.go
@@ -91,7 +91,7 @@ type OWASPCRS struct {
 	// ParanoiaLevels specifies the OWASP ModSecurity Core Rule Set (CRS)
 	// paranoia levels to use.
 	//
-	// +kubebuilder:default={}
+	// +kubebuilder:default={blocking:1,detection:1}
 	ParanoiaLevels ParanoiaLevels `json:"paranoiaLevels,omitempty"`
 
 	// ScoreThresholds specifies the OWASP ModSecurity Core Rule Set (CRS)
--- a/config/crd/bases/networking.datumapis.com_trafficprotectionpolicies.yaml
+++ b/config/crd/bases/networking.datumapis.com_trafficprotectionpolicies.yaml
@@ -67,7 +67,9 @@ spec:
                         Core Rule Set (CRS).
                       properties:
                         paranoiaLevels:
-                          default: {}
+                          default:
+                            blocking: 1
+                            detection: 1
                           description: |-
                             ParanoiaLevels specifies the OWASP ModSecurity Core Rule Set (CRS)
                             paranoia levels to use.
--- a/test/crd/crd_defaults_test.go
+++ b/test/crd/crd_defaults_test.go
@@ -1,0 +1,139 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+// Package crd runs the generated CRDs against a real apiserver (envtest) to
+// guard invariants the fake client cannot enforce. The apiserver validates a
+// schema node's structural default against that node's own CEL rules at CRD
+// registration, so an invalid default is rejected here — the fake client used
+// by the rest of the suite never sees schema, defaults, or CEL.
+//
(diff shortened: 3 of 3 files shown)


yahyafakhroji (approved):

aflor024 (approved):
Approved

PR #262, opened 2026-07-13
docs: delineate cert health-gating from re-issuance

### LINKED ISSUE #260: Unify or delineate cert-failure recovery: re-issuance (#259) vs health-gating (#212)
## Context

Two independent cert-failure-handling paths now live in the gateway controller:

- **Health-gating** (`evaluateListenerCertHealth` / `listenerCertHealth`, from #212): stop serving / isolate a bad cert so one tenant's failed cert can't freeze a shared listener. *Defensive — don't propagate the failure.*
- **Fast-track re-issuance** (`reissueFailedCertificate`, from #259 / originally #124): delete a `Certificate` stuck failed to trigger GC of the chain and recreate fresh, bypassing cert-manager's exponential backoff. *Active — recover from the failure faster.*

They don't conflict mechanically and #259 keeps re-issuance standalone (diff maps 1:1 to #124). But they overlap in intent and both key off "cert is failing."

## Question

Should the two be **unified** (re-issuance driven from the `listenerCertHealth` result instead of its own failure detection) or **left delineated** with a documented boundary?

## Considerations

- Single source of truth for "cert is failing" vs. two detectors that can disagree.
- Ordering / interaction: does deleting a cert for re-issuance perturb health-gating state or metrics?
- Re-issuance mutates state (delete + annotation bump); health-gating is read/report. Mixing mutation into the health path needs care.
- Bounded ACME retries (`maxRetries`) must still hold if triggered from the health flow.

## Acceptance

- Decision recorded: unify or keep separate.
- If unified: re-issuance consumes `listenerCertHealth`; single failure detector; tests cover the combined path.
- If separate: doc comment at both call sites stating the boundary and why.

## Related
- #259 (re-issuance, current)
- #124 (original re-issuance PR, superseded by #259)
- #212 (health-gating)
- #94 (delete errored Challenge — predecessor)

ecv:
## What

Records the boundary between the two cert-failure paths in the gateway controller with a short comment at each call site. No behavior change — comments only.

- **Health-gating** (`evaluateListenerCertHealth`, call site ~line 279): read-only, gates a listener on *any* unusable cert (`Ready!=True`, not-yet-valid, expired, missing secret, corrupt keypair).
- **Fast-track re-issuance** (`reissueFailedCertificate`, call site ~line 950): mutates (delete + gateway annotation bump), fires *only* on cert-manager hard-fail (`LastFailureTime`). Expired/missing certs recover via renewal or the ensure loop's `isNew` path instead.

## Why

Resolves the decision in #260 — keep the two paths delineated rather than unified behind a single failure detector. Full rationale and the stuck-state verification are recorded on that issue. Summary:

- The two paths key off **different signals by design**; only the narrow `LastFailureTime` signal should trigger the destructive delete-recreate.
- Health stays read-only (drives customer-facing status + metrics); re-issuance stays the sole mutator.
- Verified no truly-stuck state exists within NSO's model at the pinned cert-manager v1.17.2 (every unhealthy branch has a recovery owner).

## Test

`go build` + relevant `internal/controller` tests pass; no logic touched.

Closes #260

### DIFF
--- a/internal/controller/gateway_controller.go
+++ b/internal/controller/gateway_controller.go
@@ -276,6 +276,11 @@ func (r *GatewayReconciler) ensureDownstreamGateway(
 
 	// Evaluate certificate health once so the listener we propagate and the
 	// status we report to the user can never disagree.
+	//
+	// Deliberately separate from fast-track re-issuance (see
+	// reissueFailedCertificate): this path is read-only and gates on any
+	// unusable cert, re-issuance mutates and fires only on cert-manager
+	// hard-fail (LastFailureTime). See #260.
 	listenerCertHealth := r.evaluateListenerCertHealth(
 		ctx,
 		downstreamClient,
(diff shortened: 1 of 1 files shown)


ecv:
wow, its me suggesting a comment

scotwells (approved):

Issue #242, opened 2026-07-09
TrafficProtectionPolicy Enforce mode: 500 with truncated/empty body on all traffic


kevwilliams:
## Summary

`TrafficProtectionPolicy` (TPP) in `Enforce` mode causes a **100% reproducible 500** with a truncated response body, on a plain HTTPProxy with no unusual traffic (small static page, `GET /`, no request body). Disabling the TPP fixes it immediately.

## Environment

- Project: `default-ec6p8s`
- Proxy: `kev1n-proxy` (`HTTPProxy` -> `Gateway`/`HTTPRoute` -> edge hostname `52f1481fa40341799b03b2bd02432592.datumproxy.net`)
- Backend: `https://kev1n-org.vercel.app` (static Vercel-hosted site, 2408-byte HTML response)
- `TrafficProtectionPolicy` spec:
  ```yaml
  apiVersion: networking.datumapis.com/v1alpha
  kind: TrafficProtectionPolicy
  spec:
    mode: Enforce
    ruleSets:
    - type: OWASPCoreRuleSet
      owaspCoreRuleSet:
        paranoiaLevels: {detection: 1, blocking: 2}
        scoreThresholds: {inbound: 5, outbound: 4}
    targetRefs:
    - group: gateway.networking.k8s.io
      kind: Gateway
      name: kev1n-proxy
  ```
- TPP status: `Accepted=True` (policy is correctly attached, not a translation/acceptance failure)

## Repro

```
$ curl -v https://52f1481fa40341799b03b2bd02432592.datumproxy.net/
< HTTP/1.1 500 Internal Server Error
< content-length: 13001
< content-type: text/html; charset=UTF-8
<
* TLS alert, close notify
* end of response with 13001 bytes missing
curl: (18) end of response with 13001 bytes missing
```

- 100% reproducible across 5+ consecutive requests.
- Direct-to-origin (`curl https://kev1n-org.vercel.app/`) returns a clean `200`, 2408-byte body.
- `Content-Length: 13001` does not match the origin's real body size (2408 bytes) — it's suspected to be the size of the NSO extension server's branded 5xx error page (`errorPage.bodyPath: /etc/datum/error-pages/error-5xx.html` in the Coraza config), but **zero bytes of any body are actually delivered** before the connection closes.
- Disabling the TPP (deleting/detaching it from the Gateway) immediately restores normal `200` responses with the full 2408-byte body.

## Isolation performed

[...]

PR #268, opened 2026-07-13
fix: install downstream WAF data plane in e2e CI (#263)

### LINKED ISSUE #263: E2E gating test red: turn trafficprotectionpolicy-enforce-serves-traffic green
## Summary

The E2E gating test `chainsaw/trafficprotectionpolicy-enforce-serves-traffic` (added in #245) went red on `main` — failing on **every** open PR, so the whole `E2E Tests` workflow was red across the board. This ticket tracks turning it green.

> **Update 2026-07-13 (post-#252).** The `E2E Tests` run on the #252 merge commit (`411d117c`, run `29267425732`) shows the gating test is **still red**, and #252 is **not** the unblock. The e2e runs in-cluster from the merged code with a pinned coraza image, so that run already contained #252 — yet:
> - `trafficprotectionpolicy-paranoia-validation` → **PASS** (the new #251/#252 inverted-paranoia guard works)
> - `trafficprotectionpolicy-enforce-serves-traffic` → **FAIL** — Gateway `waf-gw` never programs the Enforce WAF listener (`status: {}`, `Programmed=Unknown`), 30s assert timeout.
>
> The remaining blocker is **#242** (core Enforce filter defect), now marked as a `blocked_by` dependency. #252 still matters for **prod** (stops tenant inverted-paranoia policies 500ing), but it is orthogonal to greening this e2e.

## Failure signature

The downstream Gateway never reaches `Programmed=True`; Envoy rejects the Coraza WAF Enforce listener:

```
status.(conditions[?type == 'Programmed'])[0].status: Invalid value: "Unknown": Expected value: "True"
--- expected
-  - status: "True"
+status: {}
```

Two independent TPP-Enforce paths produced the same "500 / broken listener" family that reddens this test:

1. **Coraza filter defect** (#242) — the WAF Enforce filter breaks the listener, so the Gateway never programs (the current, remaining blocker).
2. **Inverted paranoia** (#250) — blocking paranoia level > detection level trips CRS rule **901500**, fail-closed 500 at phase 1. Fixed and now guarded (below).

## What's landed

**NSO:**
- #244 — `%`-escape the branded error page (operator allowlist + startup validation) — fixes #243 (merged 2026-07-10)
- #249 — clone `RouteBaseDirectives` before per-policy append — fixes #246 (merged 2026-07-10)
- #251 — reject inverted paranoia at admission (CEL, fail-closed) — fixes #250 (merged 2026-07-11)
- #252 — flag inverted paranoia via status condition (`Accepted=False`/`Invalid`, skip attachment) — merged 2026-07-13 16:41. **Verified green** by the new `trafficprotectionpolicy-paranoia-validation` e2e.

**Edge coraza-waf (infra repo):**
- datum-cloud/infra#3319 — pin `coraza-waf` to `v1.3.0-alpha8` — MERGED (e2e + staging now on `v1.3.0-alpha10`)
- datum-cloud/infra#3324 — response-phase `SendLocalReply(empty body)` dropped by Envoy 1.37.x — CLOSED
- datum-cloud/infra#3329 — nil-pointer panic in `EncodeData` (shadowed err) on response-phase block — CLOSED

**e2e config:**
- coraza-waf pin in `config/dev/downstream_resources/downstream-gateway.yaml:66` bumped `v1.1.1-dev2` → `v1.3.0-alpha10`.

## Remaining to turn green

**Blocked by #242.** Even with all of the above landed, run `411d117c` proves the Enforce WAF listener still fails to program (`status: {}`). The coraza `v1.3.0-alpha10` runtime fixes (#3324/#3329) addressed response-phase 500s but not this listener-programming failure.

1. Resolve #242 — the Gateway must reach `Programmed=True` with a WAF policy in `Enforce`.
2. Re-run `E2E Tests` on `main` and confirm `trafficprotectionpolicy-enforce-serves-traffic` passes.

## Acceptance criteria

- [ ] `E2E Tests` workflow green on `main`
- [ ] `chainsaw/trafficprotectionpolicy-enforce-serves-traffic` passes with the WAF in `Enforce` mode (`Programmed=True` on the downstream Gateway)
- [x] coraza-waf image pin in e2e config no longer `v1.1.1-dev2` (now `v1.3.0-alpha10`)
- [x] inverted-paranoia guard verified in e2e (`trafficprotectionpolicy-paranoia-validation` PASS)

## Related

- **Blocker:** #242 (Enforce filter defect)
- Root bugs: #242 (filter), #250 (inverted paranoia, CLOSED)
- NSO fixes: #243, #244, #246, #249, #251, #252
- Test that gates: #245
- Parity / coverage follow-ups: datum-cloud/infra#3341, datum-cloud/infra#3408, datum-cloud/infra#3402

ecv:
## What this does

Turns the `E2E Tests` gating workflow green by installing the downstream Coraza WAF data plane on the CI infra cluster (`kind-nso-infra`).

The chainsaw test `test/e2e/trafficprotectionpolicy-enforce` asserts that Gateway `waf-gw` reaches `Programmed=True`. `waf-gw` requests `gatewayClassName: datum-downstream-gateway`, but the stock `make prepare-infra-cluster` installed only the vanilla Envoy Gateway (default controllerName) plus the NSO CRDs — nothing programmed the `datum-downstream-gateway` class, so the Gateway stayed `Programmed=Unknown` (`status: {}`) and the assert timed out on every PR.

This wires the WAF data plane into the infra-cluster setup, reusing the existing `config/extension-server` and `config/tools/envoy-gateway-downstream` manifests (previously orphaned — referenced by nothing in the Makefile/Taskfiles).

## Changes

- **`make downstream-waf-dataplane`** (new `prepare-infra-cluster` dependency): installs the downstream Envoy Gateway instance (`config/tools/envoy-gateway-downstream`, controllerName `gateway.envoyproxy.io/datum-downstream-gateway`), the extension server, and the `datum-downstream-gateway` GatewayClass + Coraza `EnvoyProxy`; also loads the operator image into `nso-infra` (new `load-image-infra`). Ordered after `cert-manager`, `envoy-gateway`, and `downstream-crds` so all required CRDs exist first.
- **`config/e2e-downstream/`** (new overlay): references the `config/extension-server` base and resolves its placeholders:
  - `pki.yaml` — a self-signed CA chain (selfSigned `ClusterIssuer` → CA `Certificate` → CA `Issuer`). The one CA issues both the extension server's TLS cert (via the cert-manager CSI driver) and Envoy Gateway's client cert; both sides trust the CA. The CA `Secret` is mounted directly for the client-CA bundle, so **trust-manager is not required** (production uses a trust-manager `Bundle`; this self-contained variant keeps the kind cluster dependency-light).
  - `waf-gateway.yaml` — the Coraza `EnvoyProxy` (NodePort 30080/30443 for the kind hostPort) + the `datum-downstream-gateway` GatewayClass.
  - `extension-server-config.yaml` — Coraza `routeBaseDirectives` and branded error page enabled (the enforce test is the #243 regression, so the error page must be exercised).
  - Deployment patch pins `replicas: 1` (single-node infra cluster) and the `ServiceMonitor` is dropped (no Prometheus operator CRDs on the bare kind cluster).
- **`config/tools/envoy-gateway-downstream/kustomization.yaml`**: resolves the `extensionManager` placeholders to the extension server FQDN (`envoy-gateway-extension-server.datum-downstream-gateway.svc.cluster.local`) and the self-signed CA `Secret`.

### GatewayClass name reconciliation

I chose the **lower-blast-radius** option: **provision `datum-downstream-gateway` in the e2e path** and leave `config/e2e/config.yaml` (`downstreamGatewayClassName: datum-downstream-gateway-e2e`) untouched.

`downstreamGatewayClassName` drives the NSO operator's upstream→downstream Gateway replication, which the `gateway` and `gateway-invalid-certificate` tests depend on — they create the `datum-downstream-gateway-e2e` class inline and expect NSO to replicate onto it. The enforce test operates entirely on `nso-infra` (it creates the Gateway/HTTPRoute/TPP directly; the extension server reads them locally) and never flows through that config, so changing it would break the two gateway tests for no benefit. The WAF data plane installed here registers `datum-downstream-gateway`, which is exactly the class the enforce test hardcodes.

## Validation

- `kustomize build config/e2e-downstream` — builds clean; verified the mTLS wiring lands (tls-ca swapped to the CA `Secret`, CSI `issuer-kind: Issuer` / `issuer-name: envoy-gateway-extension-server`, client-cert `issuerRef` repointed to the `Issuer`, `SERVER_CONFIG` set, `ServiceMonitor` removed, no leftover `placeholder-*` refs).
- `kustomize build config/extension-server` and `config/e2e` — still build clean.
- `make -n prepare-infra-cluster` — target graph is coherent and correctly ordered (cert-manager + CSI driver → stock EG with CRDs → downstream CRDs → image load → downstream EG instance → e2e-downstream overlay).

**Not validated locally (needs a CI run):** I cannot run kind here, and local `kustomize --enable-helm` is blocked by a helm v4 / kustomize incompatibility, so the downstream EG Helm render, the mTLS handshake between Envoy Gateway and the extension server, extension-server health, and `waf-gw` actually reaching `Programmed=True` + a benign 200 on the wire are confirmed only by this PR's own `E2E Tests` run. The self-signed-CA `Secret`-mount approach diverges from production's trust-manager `Bundle`; if the CI run shows the CA is not resolved the way Envoy Gateway's `certificateRef` expects, the fallback is to add trust-manager + a `Bundle` and mount the resulting ConfigMap (matching production exactly).

## Relationship to #248

Draft PR #248 ("Run edge e2e in CI against the real federated architecture") builds a production-fidelity Karmada-federated edge e2e — including a WAF data plane — but as a **separate** `test-e2e-federated.yml` workflow that runs alongside the current one, and it is **blocked on an unmerged stack (#224–#228 / #247)**. It does not turn the current `test-e2e.yml` green. This PR is the minimal fix that unblocks the presently-red gating workflow now. Once #248 lands and the single-stack `test-e2e.yml` is retired, `make downstream-waf-dataplane` and `config/e2e-downstream/` can be removed.

Fixes #263. Refs #242, #248.

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -64,3 +64,29 @@ jobs:
 
       - name: Running Test e2e
         run: make test-e2e
+
+      - name: Dump downstream WAF data plane on failure
+        if: failure()
+        run: |
+          kubectl config use-context kind-nso-infra
+          echo "::group::GatewayClasses"
+          kubectl get gatewayclass -o wide || true
+          kubectl describe gatewayclass datum-downstream-gateway || true
+          echo "::endgroup::"
+          echo "::group::EnvoyProxy"
+          kubectl -n datum-downstream-gateway get envoyproxy -o yaml || true
+          echo "::endgroup::"
+          echo "::group::datum-downstream-gateway namespace + pods"
+          kubectl get ns datum-downstream-gateway --show-labels || true
+          kubectl -n datum-downstream-gateway get pods -o wide || true
+          echo "::endgroup::"
+          echo "::group::Downstream Envoy Gateway logs"
+          kubectl -n datum-downstream-gateway logs deploy/envoy-gateway --tail=200 || true
+          echo "::endgroup::"
+          echo "::group::Extension server"
+          kubectl -n datum-downstream-gateway describe deploy envoy-gateway-extension-server || true
+          kubectl -n datum-downstream-gateway logs -l app.kubernetes.io/component=envoy-gateway-extension-server --tail=200 || true
+          echo "::endgroup::"
+          echo "::group::Certificates"
+          kubectl -n datum-downstream-gateway get certificate,secret || true
+          echo "::endgroup::"
--- a/Makefile
+++ b/Makefile
@@ -153,13 +153,18 @@ set-image-controller: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image ghcr.io/datum-cloud/network-services-operator=${IMG}
 
 .PHONY: prepare-infra-cluster
-prepare-infra-cluster: cert-manager envoy-gateway external-dns downstream-crds
+prepare-infra-cluster: cert-manager envoy-gateway external-dns downstream-crds downstream-waf-dataplane
 
 .PHONY: downstream-crds
 downstream-crds: ## Install NSO CRDs on the downstream (infra) cluster that the replicator mirrors into it.
 	$(KUBECTL) apply -f config/crd/bases/networking.datumapis.com_connectors.yaml
 	$(KUBECTL) apply -f config/crd/bases/networking.datumapis.com_httpproxies.yaml
 	$(KUBECTL) apply -f config/crd/bases/networking.datumapis.com_trafficprotectionpolicies.yaml
+
+.PHONY: downstream-waf-dataplane
+downstream-waf-dataplane: kustomize load-image-infra ## Install the Coraza WAF data plane (downstream EG instance + extension server + datum-downstream-gateway GatewayClass) on the infra cluster.
+	$(KUSTOMIZE) build --enable-helm config/tools/envoy-gateway-downstream | kubectl apply --server-side=true --force-conflicts -f -
+	$(KUSTOMIZE) build config/e2e-downstream | kubectl apply --server-side=true --force-conflicts -f -
 
 .PHONY: prepare-e2e
 prepare-e2e: chainsaw set-image-controller cert-manager load-image-all deploy-e2e
@@ -173,6 +178,10 @@ load-image-all: load-image-operator
 .PHONY: load-image-operator
 load-image-operator: docker-build kind
 	$(KIND) load docker-image $(IMG) -n nso-standard
+
+.PHONY: load-image-infra
+load-image-infra: docker-build kind
+	$(KIND) load docker-image $(IMG) -n nso-infra
 
 .PHONY: cert-manager
 cert-manager: cmctl
--- a/config/dev/downstream_resources/downstream-gateway.yaml
+++ b/config/dev/downstream_resources/downstream-gateway.yaml
@@ -67,7 +67,7 @@ spec:
             imagePullPolicy: IfNotPresent
             command:
               - cp
-              - /app/coraza-waf.so
+              - /coraza-waf.so
               - /opt/coraza-waf/
             volumeMounts:
               - name: coraza-waf
--- a/config/e2e-downstream/extension-server-config.yaml
+++ b/config/e2e-downstream/extension-server-config.yaml
@@ -1,0 +1,12 @@
+apiVersion: apiserver.config.datumapis.com/v1alpha1
+kind: NetworkServicesOperator
+gateway:
+  coraza:
+    routeBaseDirectives:
+      - Include @crs-setup-conf
+      - Include @recommended-conf
+      - SecResponseBodyLimitAction ProcessPartial
+      - SecResponseBodyLimit 32000
+  errorPage:
+    enabled: true
+    bodyPath: /etc/datum/error-pages/error-5xx.html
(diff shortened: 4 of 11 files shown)


ecv:
Heads-up while iterating on this: once the downstream WAF data plane actually installs and `waf-gw` starts programming, the enforce test stops being a pure harness check and begins exercising the real WAF **Enforce** data path. At that point, treat a **flapping** result as signal, not noise:

- **Deterministic red** (every run, same `Programmed=Unknown` / `status: {}`) = still the harness gap — the `datum-downstream-gateway` class / EnvoyProxy / extension server isn't wired yet.
- **Flap** (same commit green on some runs, red on others) = the test is likely catching a **real intermittent defect** — propagation/timing/race in Enforce programming (cf. the TPP Observe→Enforce propagation-lag family, #3402 / network-services-operator#254). Don't just re-run to green; capture the failing run's EnvoyProxy/extension-server/Envoy xDS state.

So: don't chase green blindly. A green that only holds sometimes is a bug the new coverage found, which is the point of the test.

scotwells (approved):

PR #282, opened 2026-07-15
feat(telemetry): alert on gateway reconcile-error ratio

ecv:
Rebase of #218 onto current `main` (the original was 129 commits behind), plus a one-word anchor fix. Draft for review.

## What

Two alerts on the gateway controller's reconcile error ratio, so a controller that can't apply configuration pages operators directly instead of failing silently:

- **GatewayControllerReconcileErrorRatioHigh** (warning) — error ratio > 20% for 15m
- **GatewayControllerReconcileErrorRatioCritical** (critical) — error ratio > 50% for 10m

Both link a new runbook. In production after v0.23.4 the controller ran at ~83% reconcile failure for hours with nothing firing (cert-listener withholding empties a downstream Gateway's listeners, the CRD rejects it, the controller hot-loops, stale config stays programmed). These would have fired immediately.

## Safety

- **Pure telemetry + docs, zero code** — no controller/API/RBAC/runtime change; cannot affect the operator or data plane.
- Purely additive: 3 new files + an alert group appended to the existing `gateways.yaml`.
- PromQL is valid — controller is still `Named("gateway")`; `controller_runtime_reconcile_total` is standard controller-runtime; no divide-by-zero (the error series is a subset of the total).

## Difference from #218

The original's `runbook_url` anchors had a typo — `#gatewaycontrollerreconcileerr**ato**high` / `...erratocritical` — that didn't match the runbook headers (`...ErrorRatioHigh/Critical`), so the deep-links broke. Fixed to `errorratio*` in the alert and both test fixtures.

## Verification

- `promtool check rules` — SUCCESS (2 rules); `promtool test rules` — SUCCESS (the CI `test-prometheus-rules` task auto-globs the fixture).
- `kustomize build config/telemetry/alerts` — OK.

Supersedes #218.

### DIFF
--- a/config/telemetry/alerts/gateways.yaml
+++ b/config/telemetry/alerts/gateways.yaml
@@ -50,6 +50,51 @@ spec:
         summary: "Gateway {{ $labels.resource_name }} has been degraded for over 60 seconds"
         description: "Gateway {{ $labels.resource_name }} in namespace {{ $labels.resource_namespace }} has been in a degraded state for over 60 seconds without recovering, which exceeds the 60-second SLO threshold."
 
+  # Controller reconcile-error alerts detect when the gateway controller is
+  # failing to write downstream Gateway objects. A high error ratio means updates
+  # are being rejected by the API server (e.g. invalid objects), the controller
+  # is hot-looping, and the stale configuration stays programmed at the edge.
+  - name: nso-gateway-controller-health
+    interval: 30s
+    rules:
+    # Fires when more than 20% of gateway controller reconcile attempts are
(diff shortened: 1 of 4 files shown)


scotwells (approved):
Seems like these could also be generic alerts across all controllers?

ecv:
Good call @scotwells — the same reconcile-error hot-loop is a silent-failure class across every controller, not just the gateway one. Filed #289 to generalize this into a per-controller alert class using this PR as the template.

Merging this gateway-specific pair now to close the immediate prod gap; #289 tracks promoting it to the generic class (and deciding whether the gateway rules stay as a tighter overlay or get subsumed).

Issue #283, opened 2026-07-16
Stale hostname CNAME/ALIAS not reaped on hostname removal — breaks resolution (SERVFAIL)


ecv:
## Summary

When a custom hostname is removed from an HTTPProxy/Gateway (or its proxy binding is torn down), the DNS record the platform created for it (a `CNAME` for non-apex, `ALIAS` for apex, pointing at the proxy's `*.datumproxy.net` canonical name) can be left behind. The stale record then points at a canonical name that no longer exists — so the hostname **fails to resolve** (authoritative `SERVFAIL`) and can block a later resource from re-claiming it.

> [!WARNING]
> **Confirmed via DNS, with live customer impact.** Originally filed against the HTTPProxy-**delete** path — that path's cleanup is present and fixed since NSO **v0.17.0** (prod v0.24.3). The live gap is on **hostname/binding removal**: the customer-hostname record survives after its proxy canonical is gone. Only formal step left is confirming the upstream proxy no longer lists the hostname (see _Open question_).

## How the records flow (where reap must happen)

1. **Upstream (project control plane)** — `gateway_dns_controller` creates a per-hostname `DNSRecordSet`, **Gateway-owned**. On hostname removal it should be reaped by the label-based `garbageCollectDNSRecordSets` (diff desired-vs-existing, delete stale).
2. **Downstream projection (host cluster, `ns-<uid>`)** — a mirrored copy owned by the DNSZone + an **anchor ConfigMap**. GC'd only when its upstream is deleted (anchor cascade). _Anchor-ownership here is by design, not the bug._
3. **PowerDNS** — `dns-operator` writes/removes the RRset and resolves `ALIAS` targets at query time; a live CR ⟹ the RRset is kept. If the ALIAS target is `NXDOMAIN`, PowerDNS returns `SERVFAIL` for the name.

Reaping a removed hostname therefore depends entirely on step 1 (upstream `garbageCollectDNSRecordSets`).

## Evidence (prod — customer AB1889, `ab.dk`, project `ab-generic-1qnl3i`)

- Downstream `DNSRecordSet` `ab-website-g2rfww-9f7feda5` = `ALIAS www.ab.dk → loneliness-snake-9nkbf.datumproxy.net`; `Programmed=True`, `Accepted=True`, created 2026-07-12, still present days later. Anchor present ⟹ upstream `DNSRecordSet` still exists.
[...]

PR #284, opened 2026-07-16
fix: reap DNS record when a hostname is removed

mattdjenkinson:
## Summary

Fixes the reap gap described in #283. Removing a custom hostname from a Gateway could leave its DNSRecordSet orphaned in PowerDNS, because `ensureHostnameVerification` granted a grace period to hostnames still present on the *downstream* gateway's pre-reconcile listeners, a mechanism meant to let a hostname survive its Domain being deleted after verification. That check never confirmed the hostname still existed on the *upstream* gateway at all, so a hostname removed outright stayed "claimed" for one extra reconcile pass. Nothing forces that follow-up pass to happen, so the stale record could persist indefinitely, matching the live `ab.dk` / `www.ab.dk` incident in datum-cloud/infra#3407.

The fix restricts the downstream grace period to hostnames that are still present in the upstream gateway's current listeners, closing the gap while preserving the original deleted-Domain behavior.

## Root cause

Confirmed with a fast in-process regression test rather than a full dev-cluster repro. Chaining `ensureHostnamesClaimed`, `getDesiredDownstreamGateway`, and `ensureDNSRecordSets` across two reconcile passes (hostname present, then removed) reproduces the orphan against the old code and passes against the fix.

## Changes

`ensureHostnameVerification` now only treats a downstream-listener hostname as still-verified when it's also present in the upstream gateway's current `spec.Listeners`. Added `TestHostnameRemoval_ReapsRecordInSameReconcile` in `gateway_dns_reap_regression_test.go`, which fails against the old code and passes against the fix. `go test ./...`, `go vet`, and `gofmt` are all clean.

## Not in this PR

Per the plan in #283, gateway-delete and proxy-delete DNS cleanup currently have no test coverage, and there's no chainsaw e2e test touching DNS records. Left for a follow-up, or happy to add here if preferred.

### DIFF
--- a/internal/controller/gateway_controller.go
+++ b/internal/controller/gateway_controller.go
@@ -1327,6 +1327,16 @@ func (r *GatewayReconciler) ensureHostnameVerification(
 
 	gatewayDefaultHostname := r.gatewayCanonicalHostname(upstreamGateway)
 
+	// Get a unique set of hostnames currently declared on the upstream gateway.
+	// This must be computed before the downstream grace-period handling below,
+	// since that logic only applies to hostnames the user still wants.
+	hostnames := sets.New[string]()
+	for _, l := range upstreamGateway.Spec.Listeners {
+		if l.Hostname != nil {
+			hostnames.Insert(string(*l.Hostname))
+		}
+	}
+
 	// Allow hostnames which have been successfully configured on the downstream
 	// gateway stay on the gateway, regardless of whether or not there is a
 	// matching Domain that's verified. This is done in case the user deletes the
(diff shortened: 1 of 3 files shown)


ecv:
Rebased the #283 test suite (#285) onto this branch and reran — reporting the before/after.

The suite is three tests covering the reap machinery: hostname-removal round-trip (apex ALIAS + non-apex CNAME), Gateway-deletion cleanup, and the HTTPProxy→child-Gateway owner-ref wiring the GC cascade relies on. They pass on `main` and they pass rebased on this fix. `go build`/`go vet` clean, full `internal/controller` package green, no regressions, and the two new test files coexist fine.

That green-both-ways is the useful signal: these exercise the reap mechanisms, which were already individually correct — so they confirm the real #283 defect was upstream of them, in the hostname-claim grace period this PR fixes. Matches your commit message exactly.

So the coverage splits cleanly: your regression test pins the specific bug (red on old, green here); this suite guards the surrounding plumbing (gateway-delete, proxy-delete cascade, apex/non-apex) against future regressions. Complementary, not redundant.

#285 is stacked on this branch as a draft so it rides along into the fix rather than landing on its own — merge it in or lift the file into this PR, whichever you prefer.

Still open, not covered by either: a real end-to-end/chainsaw test for the K8s GC cascade (my scenario 3 proves the pieces, not the live cascade). Worth a follow-up.

ecv (approved):

scotwells (approved):
Can we get end to end tests in place before we close the issue?

Issue #286, opened 2026-07-16
Usage metering rests on project names with no test holding them steady


ecv:
## Corrected understanding

> [!IMPORTANT]
> An earlier version of this issue claimed gateway usage metering emitted unresolvable subjects and that "all usage is rejected." **That was wrong.** Confirmed against the consumer code (`milo-os/billing`, `internal/controller/consumer`).

- The consumer attributes usage **by project name**: `consumer.go` strips `projects/` from the CloudEvent subject and matches the remainder against `BillingAccountBinding.spec.projectRef.name` (`billingaccountbinding_cache.go`).
- The NSO VRL already builds the subject as `projects/<project-name>` — **correct, no change needed**.
- `attribution_failure` fires when a project has **no active BillingAccountBinding** (or its account isn't Ready). So the high reject volume on prod is dominated by **unbilled / test projects — expected**, not a derivation bug. Accepted usage is republished to `billing.usage.<project>.valid` (no success counter), which is why "no accepted metric" *looked* like total failure but isn't.
- The production broker outage (datum-cloud/infra#3448) is a **separate** problem — event **volume** against an unsized single-replica stream, independent of attribution.

## What's actually actionable

1. **Harden the model against silent regression** — done in #287: pins the billing e2e subject to `projects/<project-name>` (it previously accepted any subject) and corrects the design doc, which called `project_name` a mere "dimension."
2. **Investigate malformed subjects.** A subset of rejected `project` values are malformed — leading-underscore `_datum-cloud`, `0-datum-internal-anchor`. These are a real (smaller) derivation/injection bug: is `idx.ProjectNames[dsNS]` returning a malformed name, or does a specific route class produce it?
3. **Confirm no billable project is wrongly rejected.** Cross-check the rejected `project` set against projects that *do* have an active BillingAccountBinding; any overlap is a real attribution bug beyond the malformed set.

ecv:
## Root cause pinpointed

The bug is in the VRL subject derivation, `config/tools/billing-usage-collector/kustomization.yaml`:

```
project_id = replace!(to_string(httproute_namespace), r'^ns-', "")   # correct — project UID from route_name
if project_name != "" {
  project_id = project_name                                          # BUG — promotes the display-name dimension to the attribution subject
}
subject = "projects/" + project_id
```

[...]

PR #287, opened 2026-07-16
test(billing): pin usage subject to project name + fix metering doc

ecv:
> [!NOTE]
> **Repurposed after review.** An earlier revision changed the VRL to build the subject from the project UID — that was wrong. The billing consumer attributes by project **name**, so the existing VRL (subject = project name) is correct and is **unchanged** here. This PR is now test + doc hardening only.

## Summary

Attribution for gateway usage runs in the billing consumer (`milo-os/billing`, `internal/controller/consumer`): it strips `projects/` from the CloudEvent subject and matches the remainder against `BillingAccountBinding.spec.projectRef.name`. So the subject must be `projects/<project-name>` — and the current VRL already does exactly that.

Two gaps let a wrong-identifier change (and a months-old misconception) look plausible. This PR closes both:

- **The e2e didn't pin the subject.** The billing e2e accepted *any* single subject that carried the four CloudEvents, so a subject built from the wrong identifier (e.g. the namespace UID) still passed. It now asserts the subject is exactly `projects/<project_name>`, and guards that `project_name` differs from the namespace UID so a UID-based regression is actually distinguishable.
- **The doc mislabeled the key.** `http-traffic-metering.md` called `project_name` a "dimension" and framed the namespace UID as the project identity — implying the subject should be the UID. Corrected: `project_name` is the attribution **subject**, with the consumer contract (subject → `projectRef.name`) now documented.

No VRL behavior change.

## Test plan

- [x] chainsaw YAML parses
- [ ] Billing e2e passes with the name-pinned subject assertion
- [ ] Confirm the assertion **fails** if the subject is built from the UID (regression guard works)

Refs #286, datum-cloud/infra#3444.

### DIFF
--- a/docs/enhancements/http-metering/http-traffic-metering.md
+++ b/docs/enhancements/http-metering/http-traffic-metering.md
@@ -144,9 +144,13 @@ central Billing System.
 
 The raw access log already carries everything the meters need *except* one
 thing: the `route_name` field identifies the owning project only by its
-control-plane namespace UID (e.g. `ns-<project-uid>`), not by the
-human-readable project name. To populate the `project_name` dimension, three
-components collaborate at the edge:
+control-plane namespace UID (e.g. `ns-<project-uid>`), not by the project name.
+This matters because the billing consumer attributes usage **by project name**:
+it strips `projects/` from the CloudEvent subject and matches the remainder
+against `BillingAccountBinding.spec.projectRef.name` (see `milo-os/billing`,
+`internal/controller/consumer`). The namespace UID would never resolve, so the
+CloudEvent subject **must** be `projects/<project-name>`. To surface the project
+name at the edge, three components collaborate:
 
 1. **Extension Server (xDS mutation).** The NSO extension server implements
    `ApplyTPPRouteConfig` in `internal/extensionserver/mutate/tpp.go`. During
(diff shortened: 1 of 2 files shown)


ecv:
## e2e verified — both directions

Ran `E2E Tests` on the fix branch and on a throwaway control branch (this fix's hardened e2e applied on top of the **pre-fix** VRL):

| Branch | `billing-http-metering` |
|---|---|
| `fix/billing-usage-project-subject` (this PR) — [run](https://github.com/datum-cloud/network-services-operator/actions/runs/29519171415) | **PASS** (84.6s) |
| `test/e2e-catches-regression` (pre-fix VRL) — [run](https://github.com/datum-cloud/network-services-operator/actions/runs/29519441377) | **FAIL** (61.6s) |

So the fix makes the billing test pass, and the hardened subject assertion **fails** against the old VRL — it genuinely catches the attribution regression.

> [!NOTE]
> The fix-branch run is overall red only because of an **unrelated** test, `gateway-invalid-certificate`, which **passed** in the control run — it's flaky, not caused by this change. Billing is green.

Control branch deleted (was verification-only).

scotwells (changes requested):
We don't attribute projects by UID. It's supposed to be the project name

scotwells (changes requested):
This is leaking a lot of implementation details in the doc. The NSO repo shouldn't know about what the billing repo does internally with usage events because that information may go stale over time.

PR #301, opened 2026-07-21
fix: tolerate clusters without Lease discovery

ecv:
## Summary

On many project control planes the Connector froze: it showed offline in the UI forever even though the desktop agent was healthy, and a pod restart didn't fix it. On staging this stranded hundreds of clusters at once, and the churn from constantly retrying fed a restart loop.

The cause: these controllers refused to start at all on any control plane that doesn't offer one optional Kubernetes API they were watching.

Now they start there too — they just skip that one optional watch and recover on their own instead of freezing.

This addresses the freeze. The related restart-loop memory growth is a separate investigation, out of scope here.

## Test plan

- [x] Build, lint, and test suite pass
- [x] New test reproduces the freeze on a control plane missing that API and confirms the controller now recovers — it fails without the fix

Related to #160


kevwilliams (approved):

scotwells:
@ecv this shouldn't have actually been needed. This was an issue with Milo not advertising the Lease API on projects which we ended up fixing.

PR #303, opened 2026-07-21
fix: requeue HTTPProxy promptly on write conflict

ecv:
## Summary

A newly created tunnel took 3-4 minutes before its toggle turned green, because the controller stalled on a retry backoff at creation instead of completing the setup right away. It now retries promptly, so a fresh tunnel programs in seconds rather than minutes.

## Test plan

- [x] New regression test forces the contended write to fail once and asserts the reconcile retries promptly instead of backing off; it fails on the pre-fix code.
- [x] `go build ./...`, `make lint`, and `make test` all pass.

Related to #166

### DIFF
--- a/internal/controller/httpproxy_controller.go
+++ b/internal/controller/httpproxy_controller.go
@@ -124,6 +124,9 @@ func (r *HTTPProxyReconciler) Reconcile(ctx context.Context, req mcreconcile.Req
 	if !controllerutil.ContainsFinalizer(&httpProxy, httpProxyFinalizer) {
 		controllerutil.AddFinalizer(&httpProxy, httpProxyFinalizer)
 		if err := cl.GetClient().Update(ctx, &httpProxy); err != nil {
+			if apierrors.IsConflict(err) {
+				return ctrl.Result{RequeueAfter: retryAfterConflict}, nil
+			}
 			return ctrl.Result{}, err
 		}
 	}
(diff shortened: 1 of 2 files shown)


kevwilliams (approved):

drewr (commented):
Looks like this is instant, infinite retries? Am I reading it wrong?

ecv:
Not instant. One-second interval. Yes forever, in practice it takes a few seconds sometimes.

scotwells:
@ecv what caused the conflict in the first place? I normally try to determine the cause of the conflict and see if there's ways of preventing the conflict before going with this approach. The audit logs via `datumctl activity history` is a great way to see the history of a resource and includes the ability to diff between each change.

ecv:
Two writers touch the Gateway spec, but they don't seem to fight over content, so the 409 looks like a timing artifact rather than a double-write.

The HTTPProxy controller owns the Gateway spec via `CreateOrUpdate`. The gateway controller writes one field on top — the default listener hostname (`gateway_controller.go:159`, `prepareUpstreamGateway`) — derived from the Gateway UID, which doesn't exist until after create, so it can't live in our create or the webhook. The HTTPProxy controller reads it back off the live object before writing, so it never clobbers it. Every other field is ours alone.

If that's right, no field has two owners, and the 409 is the two writes landing in the same resourceVersion window during the creation handoff — made more frequent by the unfiltered `Watches(&Connector{})` re-firing our reconcile on each connect-time status change. That would make retry-on-conflict the right fix, the real bug being the backoff: the default limiter treated expected 409s as failures and climbed to minutes, and nil + 1s requeue resets it.

That also means consolidating to one writer isn't on the table — the UID default forces the two-step. To cut conflicts at the source, two options: a predicate on the Connector watch so status changes we don't consume don't enqueue, or server-side apply so each controller owns its own fields and they stop colliding on resourceVersion. Retry's fine for now regardless.

I'm sure I've missed something obvious.

(`activity history` shows the PUT burst but can't attribute it — both controllers are one binary under one ServiceAccount.)

scotwells:
I wonder if [Server Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) would help prevent the conflict here?

ecv:
Oh hey look at that.

PR #308, opened 2026-07-21
docs: track version pins in prod-fidelity e2e env

### LINKED ISSUE #278: Track version pins in the prod-fidelity e2e env (why + unpin criteria)
Introduced while standing up CI on the prod-fidelity env (#276, part of #275/#247). Some versions are pinned as **compat workarounds** (should be removed once upstream catches up), others are **deliberate prod-fidelity** pins (remove only when prod moves). Tracking both so no pin lingers unexamined.

## Workaround pins (remove when the cause is gone)

| Pin | Where | Why pinned | Remove when |
| --- | --- | --- | --- |
| **helm `v3.12.3`** | `.github/workflows/test-e2e.yml` (setup-helm) | kustomize v5.5.0 probes helm with `helm version -c`; the `-c` shorthand was removed in helm 3.13, so the latest helm breaks every `kustomize build --enable-helm` in `test-infra:up`. | kustomize is bumped to a release that no longer issues the `-c` probe (or otherwise works with helm ≥3.13). Then drop the `version:` and take latest helm. |
| **kustomize `v5.5.0`** | `Makefile` `KUSTOMIZE_VERSION` | Pre-existing; it is the root cause of the helm pin above. | Bump to a release that fixes the helm probe; re-run `task validate-kustomizations` + the e2e env. Unblocks removing the helm pin. |
| **kind `v0.32.0`** | `Makefile` `KIND_VERSION` (bumped from v0.27.0) | `Taskfile.test-infra.yml` hardcodes `bin/kind-v0.32.0` (to ship the prod node image); the Makefile pin was bumped so `make kind` produces that exact binary. The hardcoded path and the Makefile var can silently drift. | Make the Taskfile reference the Makefile's `KIND` var instead of a hardcoded `bin/kind-v0.32.0` path, so a single pin governs both. Then this bump is just the normal kind pin. |

## Deliberate prod-fidelity pins (remove only on prod upgrade)

| Pin | Where | Why | Remove when |
| --- | --- | --- | --- |
| Envoy Gateway `v1.7.4` + SDK `v1.8.1` | `Taskfile.test-infra.yml`, `config/tools/envoy-gateway-downstream` | EG 1.8 does not program merged gateways served through the extension manager (waf-gw stays with empty status); 1.7.4 matches production. | Production moves off EG 1.7.4. |
| kindest/node `v1.35.5` | `Taskfile.test-infra.yml` | Exact production edge node image. | Production Kubernetes upgrade. |
| coraza-waf `v1.3.0-multiarch.1` | `Taskfile.test-infra.yml` | Production WAF filter, multi-arch so it also loads on arm64 dev hosts. | Production WAF image changes. |

## Related
- #276 — CI on the prod-fidelity env (introduces the workaround pins)
- #275 — lift the e2e suite onto the env
- #247 — the arc

ecv:
Docs-only. 

Transcribes the two pin tables from #278 into `docs/testing/e2e-version-pins.md` — workaround pins (helm, kustomize, kind) and deliberate prod-fidelity pins (Envoy Gateway + SDK, kindest/node, coraza-waf) — each with its live `file:line`, why, and unpin criteria. 

Every pin verified against the repo (no drift). 

Linked from `docs/testing/README.md`.

Closes #278

### DIFF
--- a/docs/testing/README.md
+++ b/docs/testing/README.md
@@ -80,6 +80,7 @@ edge clusters that serve traffic.
 | [`test/e2e-edge/README.md`](../../test/e2e-edge/README.md) | The real-traffic guarantees, scenario by scenario |
 | [`test/parity/README.md`](../../test/parity/README.md) | The parity check that catches silently-inert configuration |
 | [`config/federation/README.md`](../../config/federation/README.md) | Fanning configuration out to edge clusters |
+| [`e2e-version-pins.md`](e2e-version-pins.md) | Version pins in the prod-fidelity env — workarounds and deliberate prod-fidelity pins, with unpin criteria |
 
 > The design rationale that led here — the original audit of test-vs-production
 > gaps and the proposals for closing them — lives in the pull requests that
--- a/docs/testing/e2e-version-pins.md
+++ b/docs/testing/e2e-version-pins.md
@@ -1,0 +1,24 @@
+# e2e version pins
+
+The prod-fidelity e2e env (`task test-infra:up`) stands up two Kind clusters — an
+upstream control plane running the operator and a downstream cluster running the
+Envoy Gateway data plane — to mirror the production edge. Some tool versions
+there are pinned as compat workarounds (drop once upstream catches up); others
+are deliberate prod-fidelity pins (drop only when production moves). Both are
+tracked here so no pin lingers unexamined. See #278.
+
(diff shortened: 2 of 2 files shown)


ecv:
Maybe the repo is a more durable place for this than Issues? – I could go either way, in my current mood.

scotwells (approved):

PR #311, opened 2026-07-22
feat: aggregate SecurityPolicy status from edge

### LINKED ISSUE #97: Propagate security policy status back from edge
## Summary

The SecurityPolicy type does not have its status condition propagated back to consumer projects resulting in users not knowing a security policy is not configured correctly. We need to define a new [ResourceInterpreterCustomization](https://karmada.io/docs/userguide/globalview/customizing-resource-interpreter/) to propagate the status back to the user. 

## Acceptance Criteria

- SecurityPolicies in projects have the status condition set correctly based on the status in edge clusters

ecv:
Adds a Karmada `ResourceInterpreterCustomization` that aggregates `gateway.envoyproxy.io/v1alpha1` SecurityPolicy status from member clusters back to the hub, so SecurityPolicies in project control planes surface their edge status condition.

SecurityPolicy uses the Gateway API `PolicyStatus` shape (`status.ancestors[].conditions`), identical to BackendTrafficPolicy, so the new block is modeled verbatim on the existing BackendTrafficPolicy interpreter and copies `status.ancestors`.

Closes #97

### DIFF
--- a/config/federation/resourceinterpreters.yaml
+++ b/config/federation/resourceinterpreters.yaml
@@ -232,3 +232,35 @@ spec:
           end
           return desiredObj
         end
+---
+apiVersion: config.karmada.io/v1alpha1
+kind: ResourceInterpreterCustomization
+metadata:
+  name: gateway.envoyproxy.io-securitypolicy
+spec:
+  target:
+    apiVersion: gateway.envoyproxy.io/v1alpha1
+    kind: SecurityPolicy
+  customizations:
+    statusAggregation:
+      luaScript: >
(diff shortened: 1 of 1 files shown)


kevwilliams (approved):
Adds a Karmada ResourceInterpreterCustomization for gateway.envoyproxy.io SecurityPolicy that aggregates status.ancestors from the first status item back to the hub. It matches the existing gateway.envoyproxy.io-backendtrafficpolicy entry in the same file line for line (same nil-guards, same statusItems[1] approach, same status.ancestors copy), extending that pattern to a new resource kind. Config-only, no Go code touched, CI is green.

Issue #194, opened 2026-06-18
OIDC SecurityPolicy with a missing clientSecret takes down the entire shared downstream gateway (EG v1.8.x); needs e2e reproduction


scotwells:
## Summary

On the shared `datum-downstream-gateway` Envoy Gateway, a **single tenant's** OIDC `SecurityPolicy` that references a **missing `clientSecret`** takes down end-user traffic for **every** tenant on the gateway. Freshly-rolled Envoy proxy pods come up with **zero active listener filter chains** and serve nothing (HTTP + HTTPS). This was a live production incident on edge cluster `us-central-1-alice` and was resolved by rolling Envoy Gateway back from v1.8.1 → v1.7.4 (infra PR datum-cloud/infra#2746, NSO PR #193).

We need an **e2e test that reproduces this failure mode** so we can (a) prevent regressions when we re-upgrade EG, and (b) cover the broader class of "one bad tenant policy poisons the shared snapshot."

## Impact

- **Severity:** high. All-tenant data-plane outage on a shared gateway, triggered by one tenant's misconfiguration (or any not-yet-propagated OIDC secret).
- The bug is **latent**: it only manifests when a proxy pod (re)starts and pulls the poisoned xDS snapshot. Existing pods keep serving their last-good config, so it can lurk until the next rollout / node drain / scale event and then cause a wide outage.

## Root cause

EG v1.8.0 ([envoyproxy/gateway#8703](https://github.com/envoyproxy/gateway/pull/8703), "oidc: native oauth2 per-route config") reworked OIDC from a per-route filter into a **listener-level** `envoy.filters.http.oauth2` filter, with per-route config in `typed_per_filter_config`.

When the OIDC `clientSecret` is missing:
- EG correctly marks the `SecurityPolicy` `Accepted=False, reason=Invalid` ("OIDC: secret ... does not exist").
- **But it still emits the listener-level oauth2 filter with an empty/config-less `typed_config`** (just the `@type`, no `config`).
- Envoy rejects that filter with `Error adding/updating listener(s) ...: config must be present for global config`.
- Because xDS listener updates are **atomic per snapshot**, that one bad filter chain rejects the **whole** listener set (tcp-80/tcp-443/udp-443) — so every fresh proxy pod has 0 active chains.

[...]

PR #312, opened 2026-07-22
test: gate OIDC missing-secret gateway isolation

ecv:
## Summary

One tenant's broken login policy can black out the shared gateway for every tenant on it. A tenant OIDC `SecurityPolicy` pointing at a missing secret gets pushed to the shared proxy; the running proxy masks the damage by holding its last-good config, but the next restart reloads from scratch and comes up serving nothing — taking every neighbour down.

This adds the e2e test that locks in the containment: one tenant's mistake stays on its own route, even across a proxy restart. The load-bearing piece is a reusable proxy-restart step — without a forced restart the test would pass on a broken build.

**Arm 1 — bad policy never reaches the shared proxy.** OIDC policy whose secret is never created is held back (`Accepted=False, reason=PendingSecret`), nothing projected downstream, neighbour keeps serving.

**Arm 2 — from-scratch reload stays healthy.** Bad config forced directly onto the shared gateway, proxy restarted; afterwards the neighbour still serves, the bad route fails alone, listeners come up intact.

It also guards upgrades: passes on pinned Envoy Gateway v1.7.4 plus the #304 hold, goes red only if Envoy Gateway is bumped back into the still-open upstream bug. The scenario and README say so, so a green run is never misread as "upstream fixed".

> [!NOTE]
> Root cause is an open upstream Envoy Gateway bug (envoyproxy/gateway#6123). This test does not fix it — it stops a silent re-upgrade into it and proves the #304 hold works.

## Test plan

e2e-edge suite against the real shared data plane; not runnable in envtest or locally.

- [ ] `oidc-missing-secret-isolation` passes on pinned Envoy Gateway v1.7.4
- [ ] Arm 1: policy held, nothing projected downstream, neighbour serving
- [ ] Arm 2: after restart — neighbour serving, bad route fails alone, listeners intact

Related to #194

### DIFF
--- a/Taskfile.test-infra.yml
+++ b/Taskfile.test-infra.yml
@@ -414,7 +414,7 @@ tasks:
     vars:
       # Scenarios authored against this env's cluster names and downstream path.
       # Older fixtures targeting the previous cluster names are excluded here.
-      DEFAULT_SCENARIOS: extension-server-smoke waf-enforcement branded-error-page connector-offline-503 atomic-reject-isolation
+      DEFAULT_SCENARIOS: extension-server-smoke waf-enforcement branded-error-page connector-offline-503 atomic-reject-isolation oidc-missing-secret-isolation
       # CLI_ARGS (after --) wins; else SCENARIOS env; else the default set.
       SELECTED: '{{.CLI_ARGS | default .SCENARIOS | default .DEFAULT_SCENARIOS}}'
     deps:
--- a/test/e2e-edge/README.md
+++ b/test/e2e-edge/README.md
@@ -55,6 +55,17 @@ sharing the same proxy. The test introduces a genuinely bad certificate and
 confirms its listener is isolated while sibling listeners keep serving real
 traffic — the "one bad resource freezes everything" failure mode, contained.
 
+### One tenant's broken login policy can't black out the shared gateway
+A tenant OIDC policy that points at a missing secret must not drop every other
+tenant's traffic on the shared gateway. The test proves both defenses: the
+operator holds such a policy back until its secret is present (so the poison
+never reaches the proxy), and — even if the poison is placed directly on the
+gateway — a neighbor's route keeps serving after the shared proxy is restarted
+cold, while the bad route fails on its own. This doubles as the Envoy Gateway
+upgrade gate: it is green on the pinned v1.7.4 and flips red only if EG is bumped
+to v1.8.x while the upstream bug is unfixed, so "green" never gets misread as
+"the upstream bug is fixed".
+
 ## Running them
 
 The scenarios run against the production-fidelity environment:
--- a/test/e2e-edge/README.md
+++ b/test/e2e-edge/README.md
@@ -82,7 +82,8 @@ environment these assume.
 ## Layout
 
 - Scenario folders (`waf-enforcement/`, `connector-offline-503/`,
-  `branded-error-page/`, `atomic-reject-isolation/`) — one guarantee each.
+  `branded-error-page/`, `atomic-reject-isolation/`,
+  `oidc-missing-secret-isolation/`) — one guarantee each.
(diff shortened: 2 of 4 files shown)


scotwells:
@ecv should I see a failing test here?

ecv:
Oops! Yeah, there totally should be. There is now. 

At least, the test should run green, and only fail if https://github.com/envoyproxy/gateway/issues/6123 hits us.

ecv:
The scenario was never actually running. `test-infra:e2e` runs an explicit `DEFAULT_SCENARIOS` allowlist in `Taskfile.test-infra.yml`, and a new folder is inert until it's named there — so the green Federated E2E check on this PR carried no signal for the new test at all. Registered it (d2bac5a) and added the folder to the suite README's layout list.

First real run came back red, and it's a genuine find rather than the upgrade gate tripping: Arm 1 times out waiting for `Accepted=False, reason=PendingSecret`. The hold itself works — the operator logs `holding SecurityPolicy: referenced secret not present downstream` and never projects the policy downstream — but it can't record that on the tenant's resource, because the manager role is missing `securitypolicies/status`:

```
failed to record held status on SecurityPolicy chainsaw-legal-aardvark/tenant-oidc: securitypolicies.gateway.envoyproxy.io "tenant-oidc" is forbidden: User "system:serviceaccount:network-services-operator-system:network-services-operator-controller-manager" cannot update resource "securitypolicies/status" in API group "gateway.envoyproxy.io" in the namespace "chainsaw-legal-aardvark"
```

Filed as https://github.com/datum-cloud/network-services-operator/issues/322. Arm 2 never gets to run, so the EG v1.7.4 upgrade gate is still unverified — moved this to draft until #322 lands.

kevwilliams (approved):
Mirrors the atomic-reject-isolation scenario's structure and shared _steps conventions. Both arms assert real isolation (neighbour 200/403, bad route explicitly requires 5xx and fails on 200, listener filter-chain dump, zero lds.update_rejected) rather than a vacuous no-crash check, and the restart step correctly forces the cold xDS pull needed to observe the regression. Good catch folding in the DEFAULT_SCENARIOS registration fix so the scenario actually runs in CI.

ecv:
Merged main in. That picked up the `securitypolicies/status` RBAC grant from #324, the original red: the guard held the policy correctly but could not write the status the assert waited on.

Arm 2 ran for the first time today. Its registration landed after the last green run, so every earlier pass had skipped this scenario.

It then failed on its own tooling, not on the behaviour under test. Three steps read the Envoy admin API by running curl inside the proxy container, which has no curl. All three parsed the exec error as if it were admin output. One went red. Two parsed the empty result as a zero counter and passed, so they could report a healthy data plane without ever reading it.

Those steps now share a port-forward helper and fail loudly when the admin API is unreachable.

Green now, on real values: `filter_chains=4 neighbour-refs=4`, `lds.update_rejected=0`, bad route 500, neighbour 200 and 403.

One caveat: the `atomic-reject-isolation` counter check stays gated behind `nackTestEnabled: false`, so CI does not exercise its fix.

kevwilliams (approved):
Port-forward helper (envoy-admin.sh) is correctly scoped: dynamic local port parsed from the port-forward log, and every exit path (timeout, curl failure, success) kills and waits on the background pid and removes the temp log, so no leaked processes/ports across the 24-attempt retry loops. The two-commit split is a real fix, not a flaky-test bandaid: commit 1 only patched the Arm 2 listener-chain check; commit 2 correctly recognized the same distroless-container-has-no-curl bug also silently broke the two update_rejected counter checks (parsing an exec error as a zero count) and extended the fix to all three callers via the shared helper. Confirmed the main merge landed the securitypolicies/status RBAC grant intact in config/rbac/role.yaml, so the core Arm 1/Arm 2 isolation logic still holds. Good to go.

ecv:
necro'd a change i made and forgot about in july, it's good to go now

PR #316, opened 2026-07-23
fix: Stop gateway DNS dual-write for v4/v6 aliases

### LINKED ISSUE #315: Gateway DNS dual-writes CNAME and A/AAAA for v4./v6. — PowerDNS 422 hot-loop
## Summary

For gateway `v4.` / `v6.` hostnames, NSO ends up with two writers on the same FQDN: Gateway DNS creates CNAMEs to the apex, while another path (ExternalDNS off gateway addresses) creates A/AAAA. PowerDNS rejects the CNAME (`422 Conflicts with pre-existing RRset`), and `dnsrecordset-powerdns` retries forever.

Staging hit this on project `datum-cloud` (`claude-test-connector-proxy`, `customer-api-l06tj1`). Live resolution still works via the A/AAAA records; the damage is a stuck reconcile loop that trips `ControllerReconcileErrorRatioCritical` for `dnsrecordset-powerdns` (alert coverage expanded in #306 / NSO alerts applied 2026-07-22).

## How we got here (staging)

1. Both CNAME and A/AAAA existed for the same `v4.` / `v6.` names (messy dual ownership; CNAME side already failing).
2. **2026-07-10 ~11:35 UTC** — Domain RDAP refresh for `prism.staging.env.datum.net` briefly set `status.nameservers` to parent `*.datumdomains.net` (no overlap with DNSZone `*.datum-staging.net`). `HasDNSAuthority` failed, so `garbageCollectDNSRecordSets` deleted the four CNAMEs even though Gateway listeners/addresses had not changed.
3. **2026-07-11 ~14:54 UTC** — next RDAP refresh restored the correct NS. Authority returned; NSO recreated the CNAMEs into the still-present A/AAAAs. Hot loop continues.

Actors for the Jul 10–11 delete/recreate were controllers only (`control@networking.datumapis.com`). No human edited the DNSRecordSets in that window.

## Root cause

Two bugs that compound:

1. **Dual-write.** Gateway DNS (`gateway_dns_controller.go`) programs CNAMEs for claimed hostnames including `v4.` / `v6.` from status addresses. ExternalDNS (or the A/AAAA path feeding it) also programs those same names as address records. Conflict detection only flags same-hostname DNSRecordSets with a *different* `managed-by`, so same-FQDN A/AAAA from the other path do not block CNAME create.
2. **GC on a single authority miss.** `ensureDNSRecordSets` omits hostnames from `desiredRecordSetNames` when `HasDNSAuthority` is false, then GC deletes them. A transient Domain NS flap is enough to reap records the Gateway still wants.

dns-operator surfaces this as repeated 422s (related: datum-cloud/dns-operator#56 for cross-type conflict handling). Stopping the dual-write and softening GC is the NSO-side prevention.

## Acceptance

- Exactly one writer owns `v4.` / `v6.` gateway hostnames (CNAME-to-apex **or** A/AAAA, not both). Prefer documenting which path is canonical and disabling the other.
- Conflict detection treats any conflicting RRset / DNSRecordSet at the same FQDN as a blocker (or clears it deliberately), not only differently-labeled peers.
- Transient Domain authority loss does not delete Gateway DNSRecordSets on a single failed check (sustained failure / backoff, or pause programming without reap).
- Unit or e2e coverage: recreate-after-authority-restore does not produce CNAME+A/AAAA at the same name; optional case that a Domain NS blip does not GC wanted CNAMEs.

## Related

- Alert that pages this: NSO #306 (generic `ControllerReconcileErrorRatioCritical`)
- dns-operator#56 — PowerDNS 422 when writing over a pre-existing RRset (defense in depth, not the dual-write fix)
- dns-operator#59 — queue saturation (separate; this incident is error ratio from 422 retries, not wait latency)

mattdjenkinson:
## Summary

Gateway `v4.` / `v6.` hostnames were getting two writers: ExternalDNS A/AAAA via the downstream DNSEndpoint, and Gateway DNS CNAMEs to the apex. PowerDNS rejects that mix with a 422, and `dnsrecordset-powerdns` retries forever. Staging hit this on project `datum-cloud` after a Domain NS flap deleted then recreated the CNAMEs into still-present A/AAAA records (see #315).

This makes Gateway DNS the only writer for those aliases. ExternalDNS keeps the apex A/AAAA only. When Domain verification or DNS authority is temporarily missing, we still keep the hostname in the desired set so GC does not delete a record the Gateway still wants. Conflict detection now treats any other managed DNSRecordSet at the same FQDN/zone as a blocker, not only a different `managed-by`.

## Test plan

- [x] Unit: authority miss retains an existing CNAME
- [x] Unit: same-FQDN peer with platform `managed-by` surfaces Conflict (no second record)
- [x] Unit: `v4.` / `v6.` claimed hostnames create CNAMEs to the canonical hostname
- [x] Unit: `gatewayDNSEndpointEntries` emits apex A/AAAA only
- [x] `go test ./internal/controller/ -run 'DNS|GarbageCollectDNS|Reap'`
- [ ] Gateway chainsaw e2e: DNSEndpoint assert is apex-only; status still lists `v4.` / `v6.` Hostname addresses

Fixes #315

### DIFF
--- a/internal/controller/gateway_controller.go
+++ b/internal/controller/gateway_controller.go
@@ -1506,6 +1506,43 @@ func managedGatewayHostnameFromStatus(
 	return ""
 }
 
+// isIPFamilyHostname reports whether hostname is a platform v4./v6. alias of
+// the gateway canonical name. Those aliases are programmed as CNAME/ALIAS
+// DNSRecordSets by Gateway DNS; ExternalDNS must not also emit A/AAAA for them
+// or PowerDNS rejects the CNAME with a 422 conflict.
+func isIPFamilyHostname(hostname string) bool {
+	return strings.HasPrefix(hostname, "v4.") || strings.HasPrefix(hostname, "v6.")
+}
+
+// gatewayDNSEndpointEntries builds ExternalDNS endpoint entries for the apex
+// (and any other non-IP-family) hostnames only. v4./v6. aliases are omitted so
+// Gateway DNS remains the sole writer for those names.
+func gatewayDNSEndpointEntries(hostnames []string, v4IPs, v6IPs []any) []any {
+	endpoints := []any{}
+	for _, hostname := range hostnames {
+		if isIPFamilyHostname(hostname) {
+			continue
+		}
+		if len(v4IPs) > 0 {
+			endpoints = append(endpoints, map[string]any{
+				"dnsName":    hostname,
+				"targets":    v4IPs,
+				"recordType": "A",
+				"recordTTL":  int64(300),
+			})
+		}
+		if len(v6IPs) > 0 {
+			endpoints = append(endpoints, map[string]any{
+				"dnsName":    hostname,
+				"targets":    v6IPs,
+				"recordType": "AAAA",
+				"recordTTL":  int64(300),
+			})
+		}
+	}
+	return endpoints
+}
+
 func (r *GatewayReconciler) ensureDownstreamGatewayDNSEndpoints(
 	ctx context.Context,
 	downstreamGateway *gatewayv1.Gateway,
@@ -1547,7 +1584,7 @@ func (r *GatewayReconciler) ensureDownstreamGatewayDNSEndpoints(
 		return result
 	}
 
-	endpoints := []any{}
+	endpoints := gatewayDNSEndpointEntries(hostnames, v4IPs, v6IPs)
 	var gatewayDNSEndpoint unstructured.Unstructured
 	gatewayDNSEndpoint.SetGroupVersionKind(schema.GroupVersionKind{
 		Group:   "externaldns.k8s.io",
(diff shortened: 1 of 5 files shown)


JoseSzycho:
I'm sorry for this one, I don't really have the knowledge to help with this review

savme (approved):

ecv:
I feel like ExternalDNS should maybe be the sole writer? Then the responsibility isn't split. 

Point a contextless Opus model at this and ask it if the implementation has more elegant alternatives. I think there's also a few code issues w/ the unverifiedDomain retain branch never being traversed and retain-on-flap retaining forever, but these may be moot if we want ExternalDNS to do this.

We might want to merge a guidance-and-reasoning doc on what bits should handle what, first, and then the code.

scotwells:
I'd agree we should make sure there's an architecture doc we're working from that captures the responsibilities of each component in the system and make sure that's well understood here before we make a change. It's helpful to rationalize these types of situations where there's dual ownership over something.

PR #318, opened 2026-07-24
fix: Use relative owner names for Gateway DNS

### LINKED ISSUE #317: Gateway DNS should write relative owner names, not FQDNs
## Summary

Gateway DNS writes `DNSRecordSet` owner names as absolute FQDNs (`help.example.com.`). Manual records and dns-operator display conventions use relative names (`help`, `@`). In the portal zone table those two styles sit side by side, so proxy-created ALIAS/CNAME rows look wrong even though DNS still works.

## Prior decision

When automatic Gateway DNS landed for AI Edge ([enhancements#615](https://github.com/datum-cloud/enhancements/issues/615)), we noticed manual records used the short form (`@`, `www`) and asked whether Gateway-managed records should do the same. We chose FQDN at the time because it removes ambiguity about which zone the owner name belongs to ([comment](https://github.com/datum-cloud/enhancements/issues/615#issuecomment-3930535411)).

That tradeoff made sense while the feature was new. In practice the ambiguity never showed up as a real problem: every Gateway `DNSRecordSet` already points at a specific `dnsZoneRef`, so the zone context is on the object. What did show up is the cost of two conventions in one table.

## Problem

`buildDesiredDNSRecordSetSpec` in `internal/controller/gateway_dns_controller.go` forces both the owner name and the CNAME/ALIAS target to absolute FQDNs (trailing dot). Targets should stay FQDN. Owner names should not.

The API allows either form ("relative to the zone or FQDN"), so this is not a programming failure. It is a convention mismatch: portal-created records look like normal relative labels; gateway-managed rows repeat the full zone name.

dns-operator already treats relative owner names as the usual shape (`www`, `@` in display/activity fixtures). Gateway-managed FQDNs fight that convention.

## Why move away from FQDN now

- **UX.** Users read the Name column as "the label in this zone." Seeing `help.mdj-test.online.` next to `@` / `www` looks broken, even when resolution is fine.
- **Consistency.** Manual creates, BIND import, and dns-operator display paths already prefer relative names. Gateway DNS is the odd one out.
- **Ambiguity is already handled.** `spec.dnsZoneRef` binds the record to a zone. Relative owner names are unambiguous in that context, which is how zone files normally work.
- **Downstream assumptions.** Helpers that match `www` or treat apex as `@` miss gateway-managed rows stored as FQDNs.

Keep absolute FQDNs for CNAME/ALIAS **content** (the target). That is where FQDN still earns its keep.

## Customer impact

For a proxy hostname like `help.mdj-test.online` in zone `mdj-test.online`, the Name column shows `help.mdj-test.online.` next to SOA/NS rows that use `@`. Users expect `help` (or `@` at apex). It reads as a bug.

## Where the behavior lives

- `internal/controller/gateway_dns_controller.go` — `buildDesiredDNSRecordSetSpec` sets `entry.Name` to the FQDN hostname
- `internal/controller/gateway_dns_controller_test.go` — asserts `wantFQDNName` values like `api.example.com.`

## Proposed fix

Write owner names relative to the matched DNSZone:

- `help.example.com` in zone `example.com` → `help`
- `a.b.example.com` → `a.b`
- apex → `@`

Keep CNAME/ALIAS **content** as absolute FQDNs with a trailing dot.

On reconcile, update existing gateway-managed records so stored names converge to the relative form. Avoid leaving a permanent mix of FQDN and relative rows. Check anything that matches on the exact `records[].name` string (status `recordSets`, conflict detection, GC) still works after the rename.

## Acceptance

- New gateway DNSRecordSets use relative owner names; apex uses `@`
- Content/targets remain absolute FQDNs
- Existing gateway-managed records converge to relative names on reconcile
- Unit tests cover subdomain, multi-label relative name, and apex
- Portal Name column shows `help` / `@` for gateway-managed rows without a display-only strip

## Related

- Prior FQDN choice: [enhancements#615 (comment)](https://github.com/datum-cloud/enhancements/issues/615#issuecomment-3930535411)
- Surfaced from cloud-portal DNS records UI review for AI Edge / proxy-created ALIAS rows

mattdjenkinson:
## Summary

Gateway DNS was writing `DNSRecordSet` owner names as absolute FQDNs (`help.example.com.`). Manual records and the portal zone table use relative labels (`help`, `@`), so proxy-created rows looked wrong next to SOA/NS even though DNS still resolved.

We originally chose FQDNs in [enhancements#615](https://github.com/datum-cloud/enhancements/issues/615#issuecomment-3930535411) to remove ambiguity. In practice the zone is already on `spec.dnsZoneRef`, and the mixed conventions are the real cost. This change writes owner names relative to the matched DNSZone (`help`, `a.b`, `@` at apex) and keeps CNAME/ALIAS **content** as absolute FQDNs. Existing gateway-managed records converge on the next reconcile.

## Testing

I tested this on staging. The record for `help` was showing as `help.mdj-test.online` before the role out and now shows as `help` post role out. I also created a new AI Edge and added a custom hostname which worked correctly. 

Screenshot 2026-07-24 at 13 46 45

## Test plan

- [x] Unit: `TestBuildDesiredDNSRecordSetSpec` / `TestEnsureDNSRecordSets` (relative name, apex `@`, multi-label, casing, FQDN convergence)
- [x] Chainsaw: `task test-infra:test-e2e -- ./test/e2e/gateway` (asserts owner name `test` and trailing-dot CNAME content)
- [x] After deploy, create/update an AI Edge hostname and confirm the portal Name column shows the subdomain (or `@`), not the FQDN

Closes #317

### DIFF
--- a/Taskfile.test-infra.yml
+++ b/Taskfile.test-infra.yml
@@ -161,10 +161,12 @@ tasks:
   dns-crds:
     desc: "Install DNSZone / DNSRecordSet / DNSZoneClass CRDs from the dns-operator Go module on the upstream cluster. Required before prepare-upstream when gateway.enableDNSIntegration is true (the operator indexes and watches these types at startup)."
     vars:
+      # Escape {{.Dir}} so Task does not treat it as a Task template var;
+      # go list must receive the literal {{.Dir}} format string.
       DNS_OPERATOR_CRD_DIR:
-        sh: go list -m -f '{{.Dir}}' go.miloapis.com/dns-operator
+        sh: go list -m -f '{{`{{.Dir}}`}}' go.miloapis.com/dns-operator
     cmds:
-      - echo "📜 installing dns-operator CRDs (upstream)"
+      - echo "📜 installing dns-operator CRDs (upstream) from {{.DNS_OPERATOR_CRD_DIR}}"
       - kubectl --context {{.UPSTREAM_CTX}} apply -f {{.DNS_OPERATOR_CRD_DIR}}/config/crd/bases/dns.networking.miloapis.com_dnszoneclasses.yaml
       - kubectl --context {{.UPSTREAM_CTX}} apply -f {{.DNS_OPERATOR_CRD_DIR}}/config/crd/bases/dns.networking.miloapis.com_dnszones.yaml
       - kubectl --context {{.UPSTREAM_CTX}} apply -f {{.DNS_OPERATOR_CRD_DIR}}/config/crd/bases/dns.networking.miloapis.com_dnsrecordsets.yaml
--- a/Taskfile.test-infra.yml
+++ b/Taskfile.test-infra.yml
@@ -99,6 +99,7 @@ tasks:
       - task: cert-manager-upstream
       - task: cert-manager-downstream
       - task: nso-image
+      - task: dns-crds
       - task: prepare-upstream
       - task: eg-downstream
       - task: extension-server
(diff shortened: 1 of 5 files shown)


savme (approved):

aflor024:
Post-merge heads up — this interacts badly with how dns-operator handles owner names.

Its queue is keyed on the raw `spec.records[].name`, and the update handler enqueues both ObjectOld and ObjectNew. So the FQDN→relative rewrite gives you two requests for one RRset: `api` writes it, `api.example.com.` finds no owner and deletes it. Both spellings qualify to the same PowerDNS name, which is exactly why the migration looks like a no-op. The delete fires in either order; whether the record survives just depends on which request lands last. When it loses, the RRset is gone but the DNSRecordSet still says `Programmed=True`, and there's no periodic requeue to fix it.

Not your bug — it's been latent in dns-operator all along, since stable owner names meant the two enqueues deduped. This just makes every gateway-managed record hit it once. Filed as datum-cloud/dns-operator#74 and fixed in datum-cloud/dns-operator#75, merged 17:12 UTC. Both orderings are covered by tests there, and both fail against unpatched main.

Timing is worth flagging though: v0.24.8 put the rewrite into prod at 10:08 UTC, and the dns-operator fix isn't there yet. The agent that runs `dnsrecordset-powerdns` had been frozen on v0.6.4 since 09:20 UTC — its Flux Kustomization was failing to apply on an env var collision and nothing surfaced it (datum-cloud/infra#3716, fixed in datum-cloud/infra#3718). Prod's agent is on v0.6.7 now, but #75 landed after v0.6.7 was cut, so it still needs a release. Until then the window where a rewrite can drop a record is open in prod. If you know of a gateway hostname that should be resolving and isn't, that's the first place I'd look.

One small thing on this side: the out-of-zone fallback in `relativeOwnerName` drops the trailing dot, so `other.example.org` becomes `other.example.org.example.com.` once dns-operator qualifies it. The old FQDN form got rejected as out-of-zone; this one gets accepted silently. Should be unreachable since you match a zone first, but `return h + "."` would keep it failing loudly.

*Edited: originally said prod was on dns-operator v0.6.6 and needed a v0.6.7 cut. v0.6.7 already existed; the real blocker was the wedged agent Kustomization in infra#3716.*

Issue #266, opened 2026-07-13
Edges serve stale WAF config for an unmeasured length of time


ecv:
## Summary

`TrafficProtectionPolicy` spec changes made on the upstream control plane take an unbounded, **unobservable** amount of time to reach the edge data plane. During that window an edge serves the **stale** (previously-valid-or-invalid) WAF config, with no signal — on the cluster or in metrics — that a given edge is behind the control plane for a given policy.

This surfaced while investigating #242: a tenant corrected an inverted-paranoia TPP upstream, but the edge continued serving the old inverted config (which 500s 100% of the route via CRS 901500 fail-closed) for a window before converging.

## Evidence

Observed on prod `us-central-1-alice`, policy `ns-c5f1cf4e…/kev1n-proxy`:

| Source | generation | `paranoiaLevels` |
|---|---|---|
| upstream (karmada control plane) | 8 | `{detection: 2, blocking: 2}` (valid — tenant-corrected) |
| edge `us-central-1-alice` (first read) | 7 | `{detection: 1, blocking: 2}` (**inverted — 500s all traffic**) |
| edge `us-central-1-alice` (later read) | 8 | `{detection: 2, blocking: 2}` (converged) |

So the projection is **eventually consistent** — it did converge — but the edge served the broken inverted config for the lag window, and nothing exposed that the edge was a generation behind. This is not a stuck/broken sync; it's propagation latency with zero visibility.

## Why it matters

- WAF config is failure-sensitive: an inverted-paranoia policy 500s 100% of a route (#242, Mechanism A). Any remediation applied upstream has a silent tail where the edge is still broken.
- With ~15+ `*-alice` edges and 500+ project control planes, there is no per-edge/per-policy way to answer "is this edge serving the current generation of this TPP?"
- The same latency applies to *any* TPP change (mode flips Observe↔Enforce, ruleSet edits), not just paranoia corrections — cf. #255 (retrigger EG translation on TPP spec change).

## Asks

1. **Observability**: expose the applied TPP generation / resourceVersion per edge (metric or status), so control-plane vs edge skew is queryable. A `spec.observedGeneration`-style signal on the edge-side attachment, or a metric labeled by policy + edge, would let us alert on lag.
2. **Bound / quantify the latency**: measure typical and worst-case upstream→edge propagation time for a TPP spec change; decide whether it needs an SLO.
[...]

PR #331, opened 2026-08-04
feat: Surface TrafficProtectionPolicy Programmed status

mattdjenkinson:
## Summary

`Accepted` on a TrafficProtectionPolicy only means the policy attached on the control plane. Tenants and e2e have no way to tell when WAF config is actually live on every edge, so Observe/Enforce can look ready while the previous generation still serves traffic.

This adds an ancestor `Programmed` condition that turns True only after the extension server successfully applies the current generation on the edge, then mirrors Karmada's AND-aggregated status upstream with M/N messaging (`Pending` / `PartialFailure` / `Programmed`). Cache and applied generation gauges plus route metadata `generation` support ops lag checks. `Accepted` stays attach-only; product labels like Protected stay in portal/CLI.

## Test plan

- [x] Unit tests for programmed-set / TPP status / controller mirroring
- [x] e2e `trafficprotectionpolicy-enforce` asserts `Programmed`
- [ ] After deploy: flip Observe→Enforce and confirm `Programmed` goes False then True, with M/N during multi-edge skew
- [ ] Confirm `Accepted` is unchanged on attach-only

## Notes for reviewers

- Needs the staging Karmada aggregation PR in `infra` before hub objects show aggregated `Programmed`.
- Ask #2 (latency SLO / alert) is still open on #266, so this does not close the issue.

Related to #266

### DIFF
--- a/api/v1alpha/trafficprotectionpolicy_types.go
+++ b/api/v1alpha/trafficprotectionpolicy_types.go
@@ -193,6 +193,10 @@ type TrafficProtectionPolicyStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=tpp
+// +kubebuilder:printcolumn:name="Mode",type=string,JSONPath=`.spec.mode`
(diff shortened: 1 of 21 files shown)


ecv (approved):
seems legit

Issue #326, opened 2026-07-29
ControllerReconcileErrorRatio flaps with unrelated traffic volume and omits the cluster


aflor024:
## What needs to happen

`ControllerReconcileErrorRatioHigh` and `ControllerReconcileErrorRatioCritical` should fire when a controller is genuinely failing to make progress, and should stay quiet otherwise. Today a controller with a steady, known set of failures flaps in and out of critical as *unrelated* healthy traffic rises and falls, and the notification doesn't say which cluster it came from.

## Why this matters

Measured on the prod infra control plane for `dnsrecordset-powerdns`, per 10-minute window:

| result | count |
| --- | --- |
| error | 193 |
| success | 604 |
| requeue | 0 |
| requeue_after | 0 |
| **total** | **797** |

Ratio at time of measurement: **24.2%**.

Six customer records cannot be written because PowerDNS rejects them as coexistence conflicts (datum-cloud/infra#3716, #315). Those retries produce a roughly constant error floor. The ratio therefore tracks the *denominator* — when legitimate DNS work is busy, 600-odd successes dilute the floor below threshold and the alert resolves; when real activity goes quiet, the same unchanged error floor reads as 96.77% or 100% and it fires again.

The result is an alert that flips on traffic patterns rather than on a change in controller health, firing and clearing roughly **120 times per 24h** into `slack-critical`. On 2026-07-29 it paged the on-call engineer seven times before being silenced, and re-fired seven minutes after the silence lapsed.

Separately, `sum by (controller)` discards `cluster`, `namespace` and `pod`. The notification names the reconciler but not where it is running, and both prod and staging fire the same alert — today the only way to tell them apart is the Alertmanager source URL.

> [!NOTE]
> An earlier version of this issue claimed the alert fires on negligible reconcile volume, citing "30 of 31 reconciles". That was wrong — it was inferred from the reported percentage rather than measured. Real volume is ~800 reconciles per 10-minute window. A minimum-volume guard would not have prevented any of the pages described above. See the comment below.

## Desired outcome

A controller with a stable set of known-failing objects does not flap in and out of critical as unrelated traffic varies. Either the alert reflects sustained inability to progress, or terminal failures stop contributing to it.

The notification identifies the cluster and namespace, so whoever is paged knows where to look without inferring it from the source link.

The promtool fixtures under `test/prometheus-rules/controllers/` cover the flapping case — a fixed error rate with a varying success rate — so a future change cannot reintroduce it.

## Related

[...]

PR #333, opened 2026-08-04
fix: stop controller error-ratio alerts flapping

aflor024:
## Summary

`ControllerReconcileErrorRatioCritical` has been firing and resolving on a loop rather than latching — roughly 100 fire/resolve pairs into `#alerts-production-critical` in 14h for a single unchanged fault, and because `severity: critical` also matches the `grafana-oncall` route, it pages on-call on every cycle. The underlying fault deserves one page, not a hundred.

Measuring prod turned up two independent flap mechanisms, not one:

- **Sparse series.** `instance-projector` reconciles a median of 4 objects per 10 minutes. 28 of 85 windows contained no increments at all, so the denominator went to zero, `0/0` dropped the series, and Alertmanager resolved — then the next reconcile restarted `for: 10m` and it fired again. It never once crossed back under the threshold (0 of 57 data-bearing windows), so this is not the ratio dipping. A 30-minute range keeps the series continuous at that volume.
- **Dilution.** A constant error rate against a variable success rate crosses the threshold in both directions as unrelated healthy traffic rises and falls, with the controller no healthier either way — the `dnsrecordset-powerdns` and staging `gateway` case already described on #326. `keep_firing_for: 30m` holds the alert through those dips.

Both changes are load-bearing and independently guarded: reverting the range fails the sparse fixture, and removing `keep_firing_for` fails the dilution fixture.

The wider range costs detection latency on a genuinely new fault: a controller that was healthy and then fails outright reaches the critical threshold in about 26 minutes instead of 16, and the warning tier moves about 4 minutes. That is the deliberate trade — on this metric a low-volume controller cannot be both instantly detectable and non-flapping.

The sparse-series half is not only a noise fix. datum-cloud/infra#3742 diagnosed the same mechanism from the opposite end: `GrafanaAlertRuleGroup`, `GrafanaContactPoint`, `GrafanaFolder` and `GrafanaNotificationPolicy` on prod reconcile about once every 15 minutes and have been failing **100% with zero successes for over 30 days**, yet `ControllerReconcileErrorRatioCritical` sits permanently in `pending` for them and has never fired — a reconcile period longer than the evaluation window means `for:` resets on every reconcile and the rule cannot latch. Every alert-rule, contact-point and notification-routing change committed to Git in that window silently never reached Grafana. The 30-minute range is what makes that condition alertable at all, so the same change that quiets a flapping controller also surfaces a month-old silent failure.

> [!WARNING]
> This raises the unalertable threshold from a 10-minute to a 30-minute reconcile period; it does not remove it. A controller reconciling less often than every 30 minutes is still invisible to this rule. The general class infra#3742 describes stays open.

Grouping now includes `cluster` and `namespace`, so a page says where the failing controller runs instead of only naming the reconciler — #326's second ask.

> [!NOTE]
> The rule still has no job or namespace selector, so it evaluates every controller-runtime binary scraped into the cluster. That is how an `nso-slo`-owned alert ends up paging about `instance-projector`, which runs in `compute-system`. Scoping it to NSO would strand the controllers above with no coverage at all, so that ownership question is left open on #326 rather than settled here.

The fault behind the current pages is 3 orphaned Instances in one project namespace (datum-cloud/compute#194) — dead data in that instance. Orphans are not inert in general: the same project-purge trigger (milo-os/milo#750, fixed by milo-os/milo#752) strands mirrored Gateways that keep serving edge traffic for a deleted project, which #338 cleans up. This PR stops the paging pattern; it fixes neither. The runbook now splits the orphan case by what the controller programs rather than calling it benign.

`task test-prometheus-rules` was not referenced by any workflow, so the fixtures this issue asks for would not have guarded anything. Added a job that runs them.

## Test plan

- [x] `task test-prometheus-rules` — all three suites pass
- [x] Sparse-series fixture fails when the range is reverted to `10m`
- [x] Dilution fixture fails when `keep_firing_for` is removed
- [x] `promtool` confirms a held alert retains its last annotations, and that the hold is bounded (resolves once `keep_firing_for` lapses)
- [x] Sparse fixture at the warning tier and a 15-minute reconcile period fixture for the infra#3742 case, both failing when the range is reverted
- [x] New CI job runs green on this PR

Related to #326

### DIFF
--- a/.github/workflows/test-prometheus-rules.yaml
+++ b/.github/workflows/test-prometheus-rules.yaml
@@ -1,0 +1,32 @@
+name: Prometheus Rules Tests
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  test-prometheus-rules:
+    name: Run on Ubuntu
+    runs-on: ubuntu-latest
+    env:
+      PROMETHEUS_VERSION: 3.13.2
+    steps:
+      - name: Clone the code
+        uses: actions/checkout@v4
+
+      - name: Install Task
+        uses: arduino/setup-task@v2
+        with:
+          version: 3.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Install promtool
+        run: |
+          curl -fsSL -o /tmp/prometheus.tar.gz \
+            "https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/prometheus-${PROMETHEUS_VERSION}.linux-amd64.tar.gz"
+          tar -xzf /tmp/prometheus.tar.gz -C /tmp
+          sudo install -m 0755 "/tmp/prometheus-${PROMETHEUS_VERSION}.linux-amd64/promtool" /usr/local/bin/promtool
+          promtool --version
+
+      - name: Running Tests
+        run: task test-prometheus-rules
--- a/config/telemetry/alerts/controller-health.yaml
+++ b/config/telemetry/alerts/controller-health.yaml
@@ -9,26 +9,36 @@ spec:
   # reconciler is failing to apply its intent. A high error ratio means writes
   # are being rejected (invalid objects, admission errors, permissions), the
   # controller is hot-looping, and the last successfully written state stays
-  # programmed. The `controller` label names the offending reconciler.
+  # programmed. The `controller` label names the offending reconciler, and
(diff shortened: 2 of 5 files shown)


0xmc (dismissed):
A very worthwhile fix

0xmc (approved):
lgtm

PR #340, opened 2026-08-07
feat: label projected namespaces with their project

### LINKED ISSUE #339: Projected namespaces do not name the project they belong to
Alerts about customer workloads on edge clusters name the affected namespace as an opaque uuid, so a responder cannot tell which project or customer is broken without a manual lookup. The operator already knows the answer at the moment it creates that namespace: it derives the name from the upstream namespace UID and labels the namespace with the upstream cluster and namespace it came from. Adding the project to those labels would close the gap where the knowledge already exists, rather than making the monitoring side reconstruct it.

## Why here

Resolving this anywhere else is harder. The uuid belongs to a namespace inside a project control plane, and those are not scraped, so nothing downstream can observe the object that would complete the join. Two other routes were considered and rejected: exporting the mapping from the platform's project resources, which do not record that UID, and scraping project control planes, which is a much larger change for the same result.

## Definition of done

- A namespace created for projected workloads carries the owning project on a label
- A responder reading an alert from an edge cluster can name the project without a lookup
- The label is present for namespaces created before the change, or a migration path is recorded

## Notes

The project ID alone would satisfy the alerting need. A display name would be better for humans but raises questions this issue does not settle — display names are user-supplied, which makes them unbounded in cardinality and potentially customer-identifying in telemetry.

Consumer side and query evidence: https://github.com/datum-cloud/infra/issues/3840

ecv:
## Summary

Alerts from edge clusters name the affected namespace as `ns-<uuid>`, so a responder cannot tell which project or customer is broken without a manual lookup. The uuid is the UID of a namespace inside a project control plane, and nothing scrapes those, so the join cannot be completed on the monitoring side.

The operator already knows the answer when it creates that namespace. Under the Milo provider the upstream cluster name *is* the project name — it is recorded today only as `meta.datumapis.com/upstream-cluster-name`, whose value is prefixed and slash-encoded and whose key does not say "project", so every consumer has to encode the cluster-name-is-project coupling itself. This stamps `resourcemanager.miloapis.com/project-name` alongside it, the same label Milo already puts on project resources, which is what [`milo_projects_info`](https://github.com/datum-cloud/infra/pull/3867#pullrequestreview-611194980) keys on as `resource_name`.

Scoping the issue narrower than it was filed: the project is not absent from these namespaces today, it is only encoded in a label that means something else. The gap this closes is a clean key, not a missing fact.

- `ProjectNameFromClusterName` returns `""` for the single-cluster provider name (`single` names a deployment cluster, not a project) and for anything that is not a valid label value (e.g. a multi-segment `org/my-project`), so the namespace write cannot be broken by a bad value — the failure mode the existing defensive guard in this function documents.
- The downstream namespace is now ensured on the update and patch paths, not only create. Namespaces projected before this change pick up the label on their next reconcile instead of needing a manual backfill, which is the third acceptance criterion. Cost is a cached namespace GET per downstream write and no write once labels match; downstream RBAC already grants the verbs.
- Project **UID** is deliberately not stamped: the multicluster key carries the name only. Display name stays out of scope, per the [decision on the consumer issue](https://github.com/datum-cloud/infra/issues/3840#issuecomment-5221301477).

Closes #339. Consumer side: [datum-cloud/infra#3840](https://github.com/datum-cloud/infra/issues/3840).

## Test plan

- [x] `make test`, `make lint`
- [x] New unit tests in `internal/downstreamclient` cover the helper, the stamped labels, the single-cluster skip, and the backfill-on-update path
- [ ] Not covered by e2e: the test env runs the single-cluster provider, where this label is intentionally absent. Verification is a `kubectl get ns -L resourcemanager.miloapis.com/project-name` on an edge cluster after rollout.

### DIFF
--- a/internal/downstreamclient/mappednamespace.go
+++ b/internal/downstreamclient/mappednamespace.go
@@ -5,12 +5,15 @@ import (
 	"fmt"
 	"strings"
 
+	resourcemanagerv1alpha1 "go.miloapis.com/milo/pkg/apis/resourcemanager/v1alpha1"
+	multiclusterproviders "go.miloapis.com/milo/pkg/multicluster-runtime"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -82,7 +85,7 @@ func (c *mappedNamespaceResourceStrategy) GetDownstreamNamespaceNameForUpstreamN
 	return fmt.Sprintf("ns-%s", namespace.UID), nil
 }
 
-func (c *mappedNamespaceResourceStrategy) ensureDownstreamNamespace(ctx context.Context, obj metav1.Object) (*corev1.Namespace, error) {
+func (c *mappedNamespaceResourceStrategy) ensureDownstreamNamespace(ctx context.Context, obj metav1.Object) error {
 	downstreamNamespace := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: obj.GetNamespace(),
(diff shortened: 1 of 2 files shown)


scotwells:
@ecv this is already in place. The cluster name label on the namespace is the upstream project

```golang
		if c.upstreamClusterName != "" {
			downstreamNamespace.Labels[UpstreamOwnerClusterNameLabel] = fmt.Sprintf("cluster-%s", strings.ReplaceAll(c.upstreamClusterName, "/", "_"))
		}
```

ecv:
Oh duh.

For future models:

The PR body already conceded it without me following the concession through. The cluster name label is the project, so this adds a second label carrying a value that already ships.

The gap I claimed was residual — that a consumer has to strip `cluster-` and undo the slash encoding — is a relabel rule, not a producer change:

```promql
label_replace(kube_namespace_labels, "project", "$1",
  "label_meta_datumapis_com_upstream_cluster_name", "cluster-_?(.*)")
```

The other half of this PR is worse. It ensured the namespace on the update and patch paths so pre-existing namespaces would pick the new label up, which costs a namespace GET on every downstream write to backfill drift that does not exist — the measurement on the consumer issue found the label on all 156 `ns-*` namespaces, resolving every firing alert.

Closing. The work moves to a relabel rule in infra, filed as a sub-issue of datum-cloud/infra#3840.

One thing surfaced that is worth keeping separately: the `DEFENSIVE GUARD` comment right above the line you quoted documents a case where `upstreamClusterName` is empty and the label is skipped entirely, because the TrafficProtectionPolicy controller's enqueue handlers do not propagate `ClusterName`. Filed as #341.

PR #343, opened 2026-08-08
docs: record what the upstream cluster name refers to

ecv:
## Summary

`meta.datumapis.com/upstream-cluster-name` is the operator's record of the owning project, and nothing in this repo said so. Under the Milo provider the multicluster key is the name of the `Project` the cluster was discovered from — a fact that lives in the milo module, so reading NSO alone leaves the label looking like cluster plumbing with no project in it.

That gap produced #339, which proposed adding a project label the operator already writes, and #340, which implemented it. Both closed. The belief was also already in tree: `extensionserver/cache` derives project names from exactly this label while carrying a TODO saying the project label is not available yet, which reads as a gap where there is none.

- `UpstreamOwnerClusterNameLabel` now documents what the name refers to per discovery mode — project name under Milo, the literal `single` otherwise — and that no project UID is available from the key.
- `UpstreamClusterNameFromLabel` points at that description. It documented the `cluster-` prefix, the slash encoding and the legacy leading-slash form in detail, and never the referent.
- The stale `ProjectNames` TODO is dropped.

Comments only, no behaviour change.

## Test plan

- [x] `go build ./...`, `gofmt`, package tests for the two touched packages

### DIFF
--- a/internal/downstreamclient/mappednamespace.go
+++ b/internal/downstreamclient/mappednamespace.go
@@ -120,6 +120,18 @@ func (c *mappedNamespaceResourceStrategy) ensureDownstreamNamespace(ctx context.
 	return downstreamNamespace, nil
 }
 
+// UpstreamOwnerClusterNameLabel carries the multicluster provider's name for
+// the upstream cluster, encoded as UpstreamClusterNameFromLabel's inverse.
+//
+// What that name refers to depends on the discovery mode (see
+// initializeClusterDiscovery in internal/cmd/manager):
+//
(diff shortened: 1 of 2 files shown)


kevwilliams (approved):
Comment-only change explaining a real hidden invariant: what UpstreamOwnerClusterNameLabel's value refers to depends on the discovery mode (a project name under the Milo provider, or the literal 'single' under the single-cluster provider), which is not inferable from the code alone. The stale TODO in cache/types.go pointing at a label that never landed is replaced with a cross-reference to the new explanation instead of restating it. No logic changed, CI is green.

